PR #6, opened 2017-07-06
added varlen_packet_tagger block

glennlive:
This block is similar to the fixed length packet tagger, except that it extracts the packet length from the first bits of the packet.  It also could be used to decode a golay24 field giving the length.


glennlive:
add copyright and improved logging

PR #7, opened 2017-07-14
variable length framer with golay and optional ASM

glennlive:
This is the first cut at the inverse of the variable-length packet tagger.  This block injects a header into a tagged stream with the length of the packet.


glennlive:
There is an issue with this implementation, so its probably not ready to pull yet.

The problem is since this block outputs more data than it consumes, the scheduler is starving the output.  I thought this could be fixed with the min_noutput_items, but it doesn't seem to work correctly.  I need the opposite of the forecast function, and although the tagged_stream block has something like that, I'm not sure how that works yet.

glennlive:
Note that this is mostly seen when you use this block to insert the sync word as well, especially if it is long (say with a preamble too).

daniestevez:
Hi Glenn,

I think that the issue is with the understanding of how flow control works for GNU Radio blocks. When the `work()` function of a block is called by the scheduler, it is asked to produce a maximum amount of output data. The `work()` function should then produce as much output as it can using the input it has currently available, but not exceeding this maximum quantity. The scheduler then passes this output to the next block in the chain.

[Recommended reading](https://www.gnuradio.org/blog/buffers/) to understand why things are organized in this manner.

Your block can't possibly work well, as it seems that it always tries to produce `packet_len + d_header_length + asm_len` or nothing at all. The scheduler might want to call your `work()` function with an output buffer that is always smaller than `packet_len + d_header_length + asm_len`.

glennlive:
So yes, as it is (or was yesterday anyway) the block certainly didn't have much success at running because the scheduler was starving the output (despite a minimum output buffer requested in the flowgraph).

And in the past, I have always attempted to produce data whenever there was space in the output buffer.  However, that often required keeping local buffers of data, and at least state and counters (like you do in the fixed tagger block).

Meanwhile, I've finally gotten around to looking deeper into tagged-streams, and ran across this:

````
    int min_output_size = calculate_output_stream_length(d_n_input_items_reqd);
    if(noutput_items < min_output_size) {
      set_min_noutput_items(min_output_size);
      return 0;
    }
    set_min_noutput_items(1);

   // WORD CALLED HERE //
````

https://github.com/gnuradio/gnuradio/blob/044b6c2bac5cff726eb90278e6156841504b796f/gnuradio-runtime/lib/tagged_stream_block.cc#L130-L134

In other words, after many years of double copying data around, it looks like it is legit to actually not produce data when there isn't enough output space.  At least that is when the min_noutput_items is set correctly (which I wasn't doing yesterday, as simply setting it in the flowgraph clearly didn't work and wouldn't be convenient anyway).

So I need to test this latest version to see if the scheduler is actually giving me the output space I want.  Otherwise, I'll have to produce some data even if I can't copy it all.  I'd like to avoid wasteful memcpys though, so maybe I'd look into the using history().

glennlive:
I've added a simple test flowgraph that seems to be working.

However, if you manually force the buffer length of the flowgraph ("Max Number of Output") to be less than the packet length, then you are correct, the block will never output correctly.

Otherwise, the block seems to work fine.

daniestevez:
Sorry for the delay. I've been on a holiday trip.

Do you deem this ready for merge now?

glennlive:
Yes, although I think switching to PDUs is going to work better for packets.

As such, my latest branch includes a PDU version of the packet tagger, which seems to work better on vector sources.  Gnuradio just doesn't handle packets as well as streams.

The new branch already includes a few other blocks as well.  The (poorly named?) "burst amble" inserts preambles and postambles around packets.  However, it doesn't yet handle timing correctly (ideally it watches over the usrp and actually knows what is transmitting).

daniestevez:
Then, should I merge this branch (varlen_framer) or the burst_amble branch or both?

glennlive:
For sure, varlen_framer (only has the one new block and has been fairly well tested) is ready to be merged.  (Even though I will probably want a PDU version soon.)

I'd like your feedback on the other branch before creating a pull request on the other blocks... although perhaps that is the idea of a pull request.  :)  I'll create one for that branch, as I would really like your thoughts better names for the blocks or if we should add other documentation, etc.  Otherwise, for what they are (lacking in QA code, etc) they are working well for me at the moment.

What isn't working well is my clock sync block.  The M&M (which I have used for years) seems to be corrupting bits when decoding from the AX100 on the bench.  But that is probably a topic for an email, and not really here in this pull request.  (Other than to say that ossman's wpcr works great on my recording, even if it is offline).

daniestevez:
I've merged varlen_framer. I'll try to take a look at burst_amble over the next week.

PR #8, opened 2017-11-10
Translated gr-kiss fsk example to gr-satellites compatible flowgraph

ijustlovemath:
Our team needed to use the fsk.grc included with gr-kiss, but since that's deprecated, I took the liberty of replacing all the gr-kiss blocks with the corresponding gr-satellites blocks. Would love if you could include this with your bundled examples, as it's one of the toughest parts for us!


daniestevez:
Thanks! It would also be helpful to port the afsk (which implements standard 1k2 AFSK packet radio) and the bpsk examples from gr-kiss.

PR #18, opened 2018-06-14
Fix bug in 1kuns image decoder

reiinakano:
Hi, I notice that if we lose an image block, we don't write out the block that we *actually* received and instead return early. I'm assuming that isn't the intended behavior?

### DIFF
--- a/python/sat_1kuns_pf_image_decoder.py
+++ b/python/sat_1kuns_pf_image_decoder.py
@@ -81,13 +81,11 @@ class sat_1kuns_pf_image_decoder(gr.basic_block):
         elif block != self.expected_block:
             # lost block
             print "Lost image block {}".format(self.expected_block)
-            self.expected_block = block + 1
-            return
         else:
             print "Received image block {}".format(block)
 
         self.f.write(data)
-        self.expected_block += 1
+        self.expected_block = block + 1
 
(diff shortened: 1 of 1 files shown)


daniestevez:
Thanks! The idea behind this is that even if we lost a single block the current JPEG image is kind of doomed. If we knew the block size (we sort of know it in the case of 1KUNS-PF, but we're not sure if it's imposed somehow by the protocol or if it could change), then we could `fseek()` over the missing data. However, your patch looks more sensible than what is currently implemented, so I'm merging it.

reiinakano:
Cool. Anyway, I dont know if this is the right place for discussion, but have you had difficulties finding a pass with a complete image? I feel like it's really rare for a single pass to capture all image packets without losing a block. I've been checking out multiple passes from satnogs but I haven't seen a single complete image so far.

daniestevez:
I've only used IQ and KISS recordings from Mike DK3WN, and they have a few missing packets but most of them are complete. Mike has a rather good station, though. From what I remember from BY70-1 and LilacSat-1, it is not so easy to get a complete JPEG image without any lost frames.

A good exercise would be to automate image reconstruction using SatNOGS data and trying to complete the image using recordings of the same pass from different stations.

PR #28, opened 2018-12-14
Drop dummy tail bits in ESEO Line decoder

mndza:
numpy.packbits pads to full bytes, so we need to throw away dummy bits first


daniestevez:
Thanks! I've merge this.

PR #56, opened 2019-10-12
explicit declare Volk library

ra1nb0w:
macOS requires explicit declaration of Volk during linking


daniestevez:
Thank you!

Could you please check if the `maint-3.8` branch (which runs with GNU Radio 3.8) also requires an explicit declaration of Volk under OSX?

ra1nb0w:
Surely, but I can't test it at moment because we are working on the transition to GR3.8
Eventually, I send the patch ;-)

Thank you

PR #57, opened 2019-10-18
Maint 3.8 kiss_submitter

K4KDR:
Added:
-notes to guide the user
-Socket PDU block set to 'TCP Client'
-option to print PDU to monitor window
-option to send PDU to GETKISS utility via TCP Server


daniestevez:
Thanks!

PR #60, opened 2019-10-21
Fix minor CCSDS_README formatting issues

nickoe:
Make links relative such that thet work on the github md parser even if
forked.

I noticed a header didn't have the correct syntax and I noticed the links where pointing to another repo, while I think it is more portable if they use relative links.


daniestevez:
Thanks, that makes a lot of sense.

Issue #167, opened 2020-09-30
32-bit OS support


jgromes:
Hi,

I'm having some issues running gr-satellites on a 32-bit OS (Ubuntu 18.04). I downloaded all the dependencies, and it does compile  and install without issues. The problem is that when I try to run the command-line program, it crashes immediately with `Segmentation fault (core dumped)` - no help text, nothing. Opening in-terminal Python session and trying `import satellites` produces segfault too, so does running in a clean environment with only PYTHONPATH variable set.

The output from running `ctest -V` is the following (only first test shown, the others fail with the exact same error). I'm not sure if this is related to the segfault, or a different issue altogether.

```
UpdateCTestConfiguration  from :/home/gromes/repos/gr-satellites/build/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/gromes/repos/gr-satellites/build/DartConfiguration.tcl
Test project /home/gromes/repos/gr-satellites/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 1
    Start 1: qa_hdlc

1: Test command: /bin/sh "/home/gromes/repos/gr-satellites/build/python/qa_hdlc_test.sh"
1: Test timeout computed to be: 10000000
1: Traceback (most recent call last):
1:   File "/usr/local/lib/python3/dist-packages/satellites/satellites_swig.py", line 14, in swig_import_helper
1:     return importlib.import_module(mname)
1:   File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
1:     return _bootstrap._gcd_import(name[level:], package, level)
1:   File "<frozen importlib._bootstrap>", line 994, in _gcd_import
1:   File "<frozen importlib._bootstrap>", line 971, in _find_and_load
1:   File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
1:   File "<frozen importlib._bootstrap>", line 658, in _load_unlocked
1:   File "<frozen importlib._bootstrap>", line 571, in module_from_spec
1:   File "<frozen importlib._bootstrap_external>", line 922, in create_module
1:   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
1: ImportError: /usr/local/lib/libgnuradio-satellites.so.3.4.0: undefined symbol: cpu_features
1: 
1: During handling of the above exception, another exception occurred:
1: 
[...]

PR #172, opened 2020-10-01
Added support for VZLUSAT-2

jgromes:
Hi,

as mentioned in #167, I added support for VZLUSAT-2, which is an upcoming 3U cubesat by Czech Aerospace Research Center, scheduled for launch with SpaceX on 18th December. The mission goal is an in-orbit technology demonstration of experimental space weather and radiation monitoring equipment, and experimental EO camera.

VZLUSAT-2 has a 9k6/4k8 FSK downlink and beacon with AX100 framing (ASM mode), IARU coordination can be found here: http://www.amsatuk.me.uk/iaru/finished_detail.php?serialnum=749. Telemetry parser includes definitions for beacon frames and data drop packets. One thing to note is that at this point we don't have a NORAD ID, so the one in the YAML file (12345) is just a placeholder. I can also provide audio recording sample if you're intersted.

Let me know if there's something that should be changed.

### DIFF
--- a/python/satyaml/VZLUSAT-2.yml
+++ b/python/satyaml/VZLUSAT-2.yml
@@ -1,0 +1,20 @@
+name: VZLUSAT-2
+norad: 12345
+data:
+  &tlm Telemetry:
+    telemetry: vzlusat_2
+transmitters:
+  9k6 FSK downlink:
+    frequency: 437.325e+6
+    modulation: FSK
+    baudrate: 9600
+    framing: AX100 ASM+Golay
+    data:
+    - *tlm
+  4k8 FSK downlink:
+    frequency: 437.325e+6
+    modulation: FSK
+    baudrate: 4800
+    framing: AX100 ASM+Golay
+    data:
+    - *tlm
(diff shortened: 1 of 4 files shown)


daniestevez:
Hi Jan,

Thanks for the PR!

The only thing that I would change is the NORAD ID, since 12345 is the ID of some spacecraft and this could cause confusion. It is best practice to use an inexistent ID, like 99999. In fact, if you are submitting this spacecraft to SatNOGS DB, I think they'll give it a temporary ID starting by 999 until it is a assigned a definitive after launching. It would be good that this matches what we have in gr-satellites, because during early orbit some people might use gr-satellites with this preliminary NORAD ID to submit telemetry to SatNOGS DB.

I could change the NORAD ID myself, as this is a small and simple detail, but please let me know if you prefer to wait to register the spacecraft in SatNOGS DB (I think you can do that now already) and enter the same in this PR.

Regarding the recording, if you want, submit a PR to the [sateliltes-recordings](https://github.com/daniestevez/satellite-recordings) repo that is short, contains one or two packets and is in WAV 48ksps 1channel int16 format (in this case, it would contain FM-demodulated data). This is not very important, because the AX100 is typical and we have several other recordings using the same mode.

jgromes:
Thanks for the tip on the NORAD ID - that's a very good point to keep it the same as the preliminary one from SatNOGS. I'll amend this PR once we have the ID, so let's keep this open until then.

PR #175, opened 2020-10-08
cmake, docs: Add ENABLE_MANPAGES option.

ryanvolz:
Add a CMAKE flag for enabling/disabling manpage generation. This defaults to OFF for Windows and ON for every other platform. Native Windows installations won't usually have a `man` reader, nor will they necessarily have `bzip2` for compressing the man pages. By having the option and defaulting to OFF on Windows, it's easier to build gr-satellites by default (e.g. for conda).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,6 +145,15 @@ else(DOXYGEN_FOUND)
 endif(DOXYGEN_FOUND)
 
 ########################################################################
+# Setup manpage option
+########################################################################
+if(NOT WIN32)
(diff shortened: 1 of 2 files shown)


daniestevez:
Thanks for this! My fault, because I didn't thought enough about Windows when adding the manpages.

I'm merging this right away to master. Do you need something to make your life easier for conda (since the current v3.5.0) won't build?

ryanvolz:
I did a quick patch to disable the man pages for Windows and restarted the conda build, then decided it would be better to have a full blown option. But that build should be fine and then with this it will work in the next release. So nothing else to be done, I think.

PR #177, opened 2020-10-10
Added support for FOSSASAT-1B/2

jgromes:
Hi,

I added support for FOSSASAT-1B and FOSSASAT-2.

FOSSASAT-1B is a revision of the FOSSASAT-1 design, keeping the 1P PocketQube format with some simplifications in the design, while FOSSASAT-2 is an entirely new design of a second-generation 2P PocketQube. The payloads include Earth observation camera, experimental ADCS and a GNSS experiment. Both of these satellites transmit CW tracking beacon and LoRa/GFSK data beacon in the UHF band, also used for telecommand and telemetry.

Since all FOSSASATs use SX1268 to transmit GFSK, this PR also includes utility blocks for SX126x/SX127x like CRC checking or packet crop (similar to how it's handled for CC11xx). Those were tested with SX1268 transceiver, in both scrambled and NRZ mode. I'm also confident this will decode SX127x FSK (since it's compatible with SX126x FSK, as long as the initial scrambler state is set to 0x1FF).

As mentioned in #167, SX126x does two bytes reflections as a part of the descrampling procedure, that can be seen in the deframer file. I wasn't sure whether to add an example of complete SX126x FSK decoding chain - let me know if that's something that should be added (on one hand it seems a bit too generic - on the other hand it's the most non-intuitive part of the decoding chain).

Also, I wanted to ask whether there are any plans to include support for LoRa transmitters via gr-lora? The frames sent by FOSSASATs over LoRa and GFSK are identical in content (in fact some packets are not supported in LoRa mode, since it would take way too long to transmit them).


python/sx12xx_remove_length.py
```diff
@@ -0,0 +1,36 @@
...
+# This file is part of gr-satellites
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+import numpy
+from gnuradio import gr
+import pmt
+
+class sx12xx_remove_length(gr.basic_block):
```
daniestevez, line 15: Isn't this block exactly the same as [`cc11xx_remove_length`](https://github.com/daniestevez/gr-satellites/blob/master/python/cc11xx_remove_length.py)?

If so, probably the name `cc11xx_remove_length` is a bit unfortunate if it's also going to serve for other chips. In fact, I see that it would be better to have a block that is a bit more general and depends on some parameters.

However, I rather not introduce a new block which is a duplicate of an existing one.

daniestevez (changes requested):
Thanks! This is a significant contribution, so I'm glad you've sent the PR. However, I think that `cc11xx_remove_length` should be used instead of adding a new block `sx12xx_remove_length`. Maybe I'm mistaken and these two do not do exactly the same thing.

daniestevez:
Thank you for this!

Reviewing your PR I realize again that we need a general infrastructure for CRC checking in gr-satellites. There is so much duplicated code in the CRC checkers, and this is all my fault. However, until something like this is in place, I think it's reasonable to add a new CRC block for the SX12xx, since I don't think any of the existing ones covers the needed functionality.

Regarding your question about LoRa satellites, for the time being I don't plan of supporting LoRa satellites in gr-satellites. Interested people can use directly gr-lora (provided it works in real situations, which still needs to be tested), and maybe add some blocks from gr-satellites to their flowgraph for telemetry parsing and telemetry submission. In the future, if this LoRa trend continues in Amateur satellites (which is to be seen, since LoRa wastes a lot of spectrum, for example) and gr-lora works well, I might add gr-lora as an optional dependency in gr-satellites and integrate it so that it can be used from SatYAML. However, gr-lora doesn't fit well into the Demodulator/Deframer scheme on which SatYAML is based, since the LoRa decoder is a single block.

python/sx12xx_remove_length.py
jgromes: As far as I can tell, most sub-GHz FSK chips follow the convention of sync-length-payload-CRC, so perhaps a more generic block to "cut out" a part of the packet based on cut start and cut length would be appropriate. The current X_remove_length implementation assumes we want to remove the first byte of the packet.

I'll replace `sx12xx_remove_length` with `cc11xx_remove_length` in this PR and open a new one to implement the more generic cut block.

daniestevez: In fact there is already [`header_remover`](https://github.com/daniestevez/gr-satellites/blob/master/python/header_remover.py), which removes any fixed number `n`of bytes from the header. Thus, `cc11xx_remove_length` is a particular case of `header_remover.py` with `n=1`, so we should also eliminate `cc11xx_remove_length`.

In fact, I think that we need a block (or two) to cut PDUs that has similar "syntax" and functionality of the `head` and `tail` UNIX functions (with the `-c` parameter, I mean). I'll open an issue to reflect this.

daniestevez:
So basically, in order not to delay unnecessarily the merging of this PR, please just remove the `sx12xx_remove_length` block. All the other more general improvements will be done afterwards. Regarding #172, the same thing, just replace the NORAD ID in the SatYAML file and it's ready to merge.

jgromes:
Hopefully good to go now, thank you once again.

One more thing I forgot to mention about the `sx12xx_check_crc` block - SX126x allows rather flexible CRC configuration, this block is only configured to the CRC FOSSASATs are using (though it is the module's default setting). When using something other than 0x1021 polynomial, 0x1D0f initial, the CRC check won't work as expected. So to follow up on your previous comment, I would say that generic CRC block is really needed.

daniestevez (approved):

daniestevez:
Thanks! Regarding the CRC, I've put up a new issue #179 to reflect this need.

Issue #203, opened 2020-12-08
warning instead of error on invalid parameters


kng:
One thing I have noted when running gr-satellites automated is the different supported parameters of it has.
Some sats/modes support --use_agc and some support --f_offset etc.
If these are used in automated scripts and they don't match, the script will not run and a pass is lost.
I would ask for a soft warning of invalid parameters instead of hard error and not running.

daniestevez:
Well, with UNIX command line tools it is standard that if the command line options are wrong then the tool will fail and do nothing, rather than give a warning and run "as best as possible". I think this is a safe policy, in that it prevents typos that might go unnoticed if the tool just prints a warning.

However, given that you're running `gr_satellites` from the SatNOGS scripts rather than manually, I see where you're coming from and I understand that this would be quite useful in your use-case. So what about adding an option like `--ignore_unknown_args` that is not on by default but when enabled does just what you say?

Now I'll just need to think how I can get [`argparse`](https://docs.python.org/3/library/argparse.html) to print warnings instead of failing when it can't parse the arguments.

kng:
Ok, I found the code for it, I understand it's only added when the fsk_demodulator is run. 
options  = parser.parse_args()
to
options, unknown = parser.parse_known_args()

PR #206, opened 2020-12-13
Warning instead of error on invalid parameters

kng:
Implements: #203

Signed-off-by: Daniel Ekman <knegge@gmail.com>

### DIFF
--- a/apps/gr_satellites
+++ b/apps/gr_satellites
@@ -44,6 +44,7 @@ def argument_parser():
 
     p.add_argument('--version', action = 'version', version = version)
     p.add_argument('--list_satellites', action = ListSatellites, nargs = 0, help = 'list supported satellites and exit')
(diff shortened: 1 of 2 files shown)


daniestevez:
Thanks!

As I commented in #203, I think it would be better to have a `--ignore_unknown_args` option that enables this kind of behaviour, rather than having it always on. Printing out unknown parameters as you've done is a good safety feature, in any case.

To have the `--ignore_unknown_args` option, I think we can first parse the arguments with `parse_known_args()` to detect whether `--ignore_unknwown_args` is present. If it's present, then we can go ahead and use the output of this call, printing the warning about unknown parameters. If  `--ignore_unknwown_args` is not present, we can parse again with `parse_args()`, which I think has the nice feature that it will print out the usage if there are unknown arguments.

kng:
I hope these changes are good, didn't know small commits would show up immediately (:

daniestevez:
Perfect, thank you very much!

I'll squash and merge your changes.

PR #246, opened 2021-03-19
Fix CubeSX-Sirius-HSE.yml carrier frequency and KSU_CubeSat.yml baudr…

bychek:
…ate. Add one more transmitter to OrbiCraft-Zorkiy.yml.

### DIFF
--- a/python/satyaml/CubeSX-Sirius-HSE.yml
+++ b/python/satyaml/CubeSX-Sirius-HSE.yml
@@ -5,7 +5,7 @@ data:
     telemetry: usp
 transmitters:
   9k6 FSK downlink:
-    frequency: 435.050e+6
+    frequency: 437.050e+6
     modulation: FSK
     baudrate: 9600
     framing: USP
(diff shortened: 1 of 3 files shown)


daniestevez:
Hi Mikhail,

Thanks for submitting these.

There is a reason why I didn't add the 2nd transmitter for OrbiCraft-Zorkiy. `gr_satellites` will instance a decoder for each transmitter and run all of them in parallel. This is intended for satellites that transmit different protocols. For the case of OrbiCraft-Zorkiy, since the protocol for both transmitters is exactly the same, if we put the two transmitters in the SatYAML file, we will get duplicated decodes, as if the packet came from both transmitters simultaneously. I think this will confuse users. Since the frequency field in the SatYAML file is currently only used for documentation, I figured out that it was best to include the primary transmitter only. Does this seem good to you?

The two other modifications fix typos I made, so thanks for fixing this. Can you modify this PR to include only these modifications?

bychek:
Sure it's ok. Done with modifying PR. Thanks for the explanation and all the hard work you did. That's for sure that we can't do it so fast as you :).

PR #251, opened 2021-03-25
Update CubeSX-Sirius-HSE.yml

janvgils:
Added the 4k8 baud rate, the satellites is switched to 4k8 as can be seen in this SatNOGS observation https://network.satnogs.org/observations/3835689/


daniestevez:
Thanks!

PR #284, opened 2021-07-03
Update TUBIN.yml

janvgils:
On the TU-Berlin website they also revere to a SIDS collector.
There I would like to suggest to add the SIDS variable.


daniestevez:
Thanks! I hadn't seen their custom SIDS server. I'm merging.

Issue #291, opened 2021-08-09
Setting fixedlen_tagger packet length while flowgraph is running


LucasBriggs154:
While using the "Sync and Create Packed PDU" block I was trying to implement setting the packet length while the flowgraph was actively running. I noticed that the "Packet length (bytes)" parameter was underlined (implying it is mutable at runtime), but it didn't seem to be working properly.

The problem lies with the Fixed Length Packet Tagger and its own packet length parameter, which is immutable at runtime. In order to enable this behavior I made some small changes to 3 files in my local gr-satellites build that I will lay out below. It amounts to creating a single callback setter for the packet_len in fixedlen_tagger.py.

I'm submitting this as an issue rather than a PR because I'm not well enough versed with gr-satellites (or gr-modules in general) to know if there's additional places that these changes need to propagate to.

Here are my changes as code blocks:
./python/fixedlen_tagger.py
```
class fixedlen_tagger(gr.basic_block):
   ...
   def set_packet_len(self, packet_len):
      self.packet_len = packet_len
   ...
```
./grc/satellites_fixedlen_tagger.block.yml
```
templates:
   ...
   callbacks:
   - set_packet_len(${packet_len})
   ...
```
./python/hier/sync_to_pdu_packed.py
```
class sync_to_pdu_packed(gr.hier_block2):
   ...
   def set_packlen(self, packlen):
      self.packlen = packlen
      self.satellites_fixedlen_tagger_0_0_0.set_packet_len(self.packlen)
```
^ I believe this last change would have been automatically added by GRC if the module was generated again from GRC with the new fixedlen_tagger callback. Should also work for the other hierarchical blocks that use the fixedlen_tagger.

I imagine the sync_to_pdu_packed.py hier block was generated with GRC, which defaults parameters to be mutable, so the fact this didn't work before makes sense. However, setting the packet length at runtime is quite useful, especially for my application, and I think that its worth including.

daniestevez:
Hi Lucas,

Thanks! I think that having the callback to set the packet length in the `fixedlen_tagger` would be a very good addition. I greatly encourage you to submit this as a PR.

I think that just the three files that you've changed are sufficient. Once thing I've spotted is that it will also be necessary to change the `maxlen` of [`self.stream`](https://github.com/daniestevez/gr-satellites/blob/master/python/fixedlen_tagger.py#L31), specially when going to a larger packet size.

[...]

PR #292, opened 2021-08-17
Add fixedlen tagger length callback

LucasBriggs154:
see: https://github.com/daniestevez/gr-satellites/issues/291

### DIFF
--- a/python/fixedlen_tagger.py
+++ b/python/fixedlen_tagger.py
@@ -38,7 +38,8 @@ class fixedlen_tagger(gr.basic_block):
 
     def set_packet_len(self, packet_len):
         self.packet_len = packet_len
-        self.stream = collections.deque(self.stream, maxlen=self.packet_len - 1)
+        self.stream = collections.deque(
+            self.stream, maxlen=self.packet_len - 1)
 
     def try_to_flush(self, out):
         # Try to send as much items as we have in buffer
--- a/python/fixedlen_tagger.py
+++ b/python/fixedlen_tagger.py
@@ -36,6 +36,10 @@ class fixedlen_tagger(gr.basic_block):
         self.written = 0
         self.really_written = 0
 
+    def set_packet_len(self, packet_len):
+        self.packet_len = packet_len
+        self.stream.maxlen = self.packet_len
+
     def try_to_flush(self, out):
         # Try to send as much items as we have in buffer
         len_write = min(len(self.data), len(out))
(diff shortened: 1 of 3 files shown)


python/fixedlen_tagger.py
```diff
@@ -36,6 +36,10 @@ def __init__(self, syncword_tag, packetlen_tag, packet_len, stream_type):
         self.written = 0
         self.really_written = 0
 
+    def set_packet_len(self, packet_len):
+        self.packet_len = packet_len
+        self.stream.maxlen = self.packet_len
```
daniestevez, line 41: This doesn't look right. According to the Python [documentation](https://docs.python.org/3/library/collections.html#deque-objects), the `maxlen` attribute of a deque is read-only. I don't know if there's a way to change the `maxlen` of a deque. Perhaps it's necessary to create a new deque and move the data.

LucasBriggs154: Ah yes, this doesn't work. Just read up on the documentation, pushing fix.

python/fixedlen_tagger.py
```diff
@@ -38,7 +38,10 @@ def __init__(self, syncword_tag, packetlen_tag, packet_len, stream_type):
 
     def set_packet_len(self, packet_len):
         self.packet_len = packet_len
-        self.stream.maxlen = self.packet_len
+        copy = self.stream.copy()
+        self.stream = collections.deque(maxlen=self.packet_len - 1)
+        for entry in copy:
+            self.stream.append(entry)
```
daniestevez, line 44: Thanks! I think this would work, but perhaps it can be made more efficient like so:
```
self.stream = collections.deque(self.stream, maxlen=self.packet_len-1)
```
The copy doesn't seem necessary (we could `old_stream = self.stream`), and instead of the `for` loop it's better to use the `.extend()` method of the deque. But actually the new deque can be constructed and initialized with an existing iterable, so it's not even necessary to use `old_stream` or `.extend()`.

LucasBriggs154: Yup, that looks much better. I believe it works just as well. Pushed another fix!

daniestevez:
Thank you! Now everything looks good to me.

There is a [pycodestyle error](https://github.com/daniestevez/gr-satellites/pull/292/checks?check_run_id=3354629840) that you should fix before I merge this. Otherwise, this is good to go. Thanks again for submitting the PR.

LucasBriggs154:
Ah, of course there is. Latest push should have taken care of the line overflow.
Thanks again for all of your help!

PR #298, opened 2021-09-16
Add support for CUTE state of health (SOH) and flight software (FSW) packets over 70cm AX.25 G3RUH

nsdecicco:
Add support to gr-satellites for the Colorado Ultraviolet Transit Experiment (CUTE):

- Add a YAML definition for CUTE, sans NORAD ID (as this is not yet known)
- Add a stateful telemetry decoder for 70cm state of health (SOH) and flight software (FSW) CUTE telemetry packets
- Make corresponding changes to CMakeLists.txt and __init__.py to include the new telemetry decoder in the build


daniestevez (changes requested):
Thanks again for this! I have read more thoroughly the code and it looks very good.

Just two things before merging:
1. Add a preliminary NORAD ID to the YAML file.
2. Take a look at the output of the [`pycodestyle` CI pipeline](https://github.com/daniestevez/gr-satellites/pull/298/checks?check_run_id=3640605636) and fix the formatting according to [PEP8](https://www.python.org/dev/peps/pep-0008/). You can run `pycodestyle` locally to check these. I think that the main problems are spaces around the `=` sign when it is used for optional parameters (this should not be done; we do `f(a, b=7, c=3)` instead of `f(a, b = 7, c = 3)`), some lines that are too long, and the presence of tabs, which can be quite problematic in Python (the standard indentation is 4 spaces).

Regarding 2., I realize that I have never written a file describing the contribution guidelines for gr-satellites. While we have `pycodestyle` running in the CI pipeline to help maintain PEP8 style, this is not immediately obvious, so I'll write now some guidelines to help other that want to contribute.

daniestevez:
Hi Nick,

Thanks for this. I've just seen in the forums that fredy is saying that they'll give the temporary NORAD ID 99510. So I think you can use that in the YAML file. It's OK if you force push the modified commit. This should retrigger CI and it should pass this time.

nsdecicco:
Hi Daniel,

Just did a force push with the temporary NORAD ID that fredy gave me--checks all passed :)

Thanks,

Nick

daniestevez (approved):

daniestevez:
Perfect! I'm merging this. Thanks again for your contribution.

Issue #300, opened 2021-09-23
Add support for the Colorado Ultraviolet Transit Experiment (CUTE) S-band transmissions


nsdecicco:
# Introduction

This is a continuation of issue #297, "Add support for the Colorado Ultraviolet Transit Experiment (CUTE)", with a focus on adding support for CUTE's S-band transmitter as opposed to its UHF transceiver.

# CUTE S-band transmitter description

Band | 13cm (S-band)
-----|-------
Mfg./Model | BCT, TX only
Frequency | 2.402 GHz
Modulation | BPSK
Baudrate | 500 kilobaud
Framing | CCSDS
Antenna | High-gain patch
Transmit power | ~2W

# Demodulation

The BPSK demodulator that currently exists in `gr-satellites` has been found to work to demodulate CUTE S-band transmissions without modification.

# Deframing

`gr-satellites` has two CCSDS deframers:

1.	CCSDS Reed-Solomon Deframer (implemented in [`python/components/deframers/ccsds_rs_deframer.py`](https://github.com/daniestevez/gr-satellites/blob/master/python/components/deframers/ccsds_rs_deframer.py))
2.	CCSDS Concatenated Deframer (implemented in [`python/components/deframers/ccsds_concatenated_deframer.py`](https://github.com/daniestevez/gr-satellites/blob/master/python/components/deframers/ccsds_concatenated_deframer.py))

Both of these assume that a forward error correcting (FEC) code is used. CUTE does not use an FEC, so neither of these blocks will work as-is. However, the former is very close to being usable as-is if the FEC block is bypassed.

The blocks internal to the CCSDS Reed-Solomon deframer are initialized in `ccsds_rs_deframer.py` [at lines 62 through 73](https://github.com/daniestevez/gr-satellites/blob/7f193ab351df8bafb50b56b3b3117c472252291b/python/components/deframers/ccsds_rs_deframer.py#L62-L73):

https://github.com/daniestevez/gr-satellites/blob/7f193ab351df8bafb50b56b3b3117c472252291b/python/components/deframers/ccsds_rs_deframer.py#L62-L73

These blocks are then connected together [on lines 75 through 86](https://github.com/daniestevez/gr-satellites/blob/7f193ab351df8bafb50b56b3b3117c472252291b/python/components/deframers/ccsds_rs_deframer.py#L75-L86):

[...]

PR #301, opened 2021-10-07
Allow disabling Reed-Solomon FEC in CCSDS deframer

c-decicco:
- Modify the ccsds_rs_deframer constructor to accept an additional argument, rs_en, which controls whether Reed-Solomon decoding is used or not.
- Do not expose rs_en to the GNU Radio end-user; instead, set rs_en to False in satellites_ccsds_rs_deframer.block.yml, and create a copy of this block, satellites_ccsds_uncoded_deframer.block.yml, with rs_en set to True.
  - Also hide Reed-Solomon related options from the user in satellites_ccsds_uncoded_deframer.block.yml
  - Add satellites_ccsds_uncoded_deframer.block.yml to the CMakeLists.txt
  - Add 'CCSDS Uncoded' to the list of accepted framings in python/satyaml/satyaml.py
  - Allow specifying 'CCSDS Uncoded' as a framing in python/core/gr_satellites_flowgraph.py

For discussion leading up to these changes, please see [issue report #300](https://github.com/daniestevez/gr-satellites/issues/300).

### DIFF
--- a/docs/source/components.rst
+++ b/docs/source/components.rst
@@ -380,16 +380,21 @@ Reed-Solomon codeword and is used by SMOG-P and ATL-1.
 CCSDS deframers
 """""""""""""""
 
-The CCSDS Concatenated deframer and CCSDS Reed-Solomon deframer blocks implement
-some of the CCSDS protocols defined in the TM Synchronization and Channel Coding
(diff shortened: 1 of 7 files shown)


daniestevez:
Thanks for this! I'm merging it.

PR #321, opened 2022-01-15
Update EXP-1.yml

janvgils:
A small error in the frequency notation.


daniestevez:
Thanks for spotting the mistake.

PR #322, opened 2022-01-15
WIP: Add support for GNU Radio 3.10

daniestevez:
This PR will add support for GNU Radio 3.10.

The idea is to make the codebase in the master branch compatible with both GNU Radio 3.9 and 3.10.


daniestevez:
it seems that some tests are flaky in the GR 3.10 Ubuntu CI. However, they run fine every time on my machine (same Dockerfile).

PR #329, opened 2022-01-30
Update GASPACS.yml

janvgils:
maybe too premature, but all indications are that GASPACS is object 51439


daniestevez:
This also appears now in Celestrak as 51493, so I think it's already quite final.

PR #335, opened 2022-02-19
Create INSPIRESat-1.yml

janvgils:
Detailed information can be found at https://github.com/Dhruva-Ananth/INSPIRESat_1_Beacon_Info


daniestevez:
Merging. There is the related discussion #332 about adding a telemetry parser for INSPIRESat-1, but it doesn't hurt to have the SatYAML file before the telemetry parser is added.

PR #337, opened 2022-02-19
Update Doxygen code from upstream gnuradio template

ryanvolz:
Since GNU Radio no longer needs `six` as a dependency, this makes it so that `gr-satellites` also no longer needs `six` just to copy over docstrings to the Python bindings. And I'm sure there are other improvements, too!


daniestevez:
Thank you! Merging.

PR #338, opened 2022-02-19
Add INSPIRESat-1 telemetry decoder

nsdecicco:
- Add a new decoder, python/telemetry/inspiresat_1.py
  - Add this decoder to the CMakeLists to include it in the build
  - Add this decoder to python/telemetry/__init__.py to make it
    available at runtime
- Add a new adapter to python/adapters.py, PolynomialAdapter
  - This adapter allows converting polynomials of arbitrarily high order,
    and is used by the INSPIRESat-1 telemetry decoder for thermistor
    temperatures.
- Update INSPIRESat-1.yml to use this new telemetry decoder

### DIFF
--- a/python/adapters.py
+++ b/python/adapters.py
@@ -31,6 +31,19 @@ class LinearAdapter(AffineAdapter):
         return AffineAdapter.__init__(self, c, 0, *args, **kwargs)
 
 
+class PolynomialAdapter(Adapter):
+    def __init__(self, coeffs, *args, **kwargs):
+        self.c = coeffs
+        return Adapter.__init__(self, *args, **kwargs)
+
+    def _encode(self, obj, context, path=None):
+        raise NotImplementedError
+
+    def _decode(self, obj, context, path=None):
+        x = float(obj)
+        return sum([c * (x ** i) for i, c in enumerate(self.c)])
+
+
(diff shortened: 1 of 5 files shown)


daniestevez:
Thanks!

I've seen that the `inspire_1.py` file doesn't have a header of the form
```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

# Copyright 2022 COPYRIGHT-HOLDER-INFO
#
# This file is part of gr-satellites
#
# SPDX-License-Identifier: GPL-3.0-or-later
#
```

Can you add this? The rest of the code seems good, so once this is in, I'll merge it.

nsdecicco:
Hi @daniestevez,

I added the header per your request. I attributed the copyright to both The Regents of the University of Colorado for the year 2022, and to you for some previous years, as a small snippet at the end of the file (where the `inspiresat_1` structure is defined) was derived from some existing gr-satellites code. (I don't recall which decoder it derived from, but it was the same that I drew from for CUTE, so I used the same copyright attribution & dates as I specified in the header of the CUTE telemetry decoder.)

Thanks, and as always, let me know if you would like anything else changed.

Nick

daniestevez:
Thanks! Now everything looks good. Merging.

PR #340, opened 2022-02-23
Fix conversions for INSPIRESat-1 telemetry decoder

nsdecicco:
The IS1 team informed me that a couple of the scale factors that were originally provided to me were incorrect: this commit aims to remedy those.

Change summary:

- sp0_curr and sp0_volt require different scalings than sp1_curr, sp1_volt, sp2_curr, sp2_volt, as sp0 corresponds to the body mounted solar panel and uses a different current/voltage measurement circuit.
- Apply a conversion to the adcs_body_rt* telemetry points
- Add higher-order terms to obc_temp, eps_temp, int_temp, sp0_temp, sp1_temp, sp2_temp telemetry points' polynomials.
- Fix bat0_temp and bat1_temp polynomials: entirely different polynomials than what is used for the other temperature telemetry points are needed for these two telementry points.

### DIFF
--- a/python/telemetry/inspiresat_1.py
+++ b/python/telemetry/inspiresat_1.py
@@ -123,6 +123,13 @@ adcs_info = BitStruct(
                                CONVERGING=5, ON_SUN=6, NOT_ACTIVE=7),
 )
 
+# Polynomial used for converting certain temperature telemetry points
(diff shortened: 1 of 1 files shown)


daniestevez:
Thanks. Merging.

PR #341, opened 2022-02-27
Update Eaglet-I.yml

janvgils:
This satellite seam to have two transmitter https://db.satnogs.org/satellite/NMSF-0466-6077-1186-9611#transmitters
Therefor I added the BPSK options to the yml file. When the FSK definition is indeed AIS then maybe the framing needs to change.

### DIFF
--- a/python/satyaml/Eaglet-I.yml
+++ b/python/satyaml/Eaglet-I.yml
@@ -3,11 +3,20 @@ norad: 43790
 data:
   &tlm Telemetry:
     telemetry: ax25
+  &tlm1 unknown:
+    unknown
 transmitters:
+  1k2 BPSK downlink:
+    frequency: 435.200e+6
+    modulation: BPSK
+    baudrate: 1200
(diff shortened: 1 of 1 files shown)


daniestevez:
Thank you. I have been checking the git history to see where the information in the current SatYAML file came from, and it turns out that I added Eaglet-I "automatically" using the procedure described in [this post](https://destevez.net/2020/01/extracting-ax-25-satellites-from-satnogs-db/) (using SatNOGS data to detect which satellites transmit AX.25). I suspect that the fact that Eaglet-I has two different transmitters could have confused my script.

I have been checking the SatNOGS observations to see if there are some examples of 9k6 FSK transmissions, but all the observations I've seen are 1k2 BPSK.

In any case, I'm merging this. It will be interesting to check if the 9k6 FSK are actually AIS. In this case gr-satellites won't be able to do anything with this transmissions (not even deframe them), so we could remove this transmitter from the SatYAML file.

PR #343, opened 2022-03-05
Fix bug in kiss_server_sink

### LINKED ISSUE #342: kiss_server_sink: constructor mismatch
self.satellites_kiss_server_sink_0 = satellites.components.datasinks.kiss_server_sink('127.0.0.1', 8100, options="")
  File "/usr/local/lib/python3.10/site-packages/satellites/components/datasinks/kiss_server_sink.py", line 38, in __init__
    self.server = blocks.socket_pdu(
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. gnuradio.blocks.blocks_python.socket_pdu(type: str, addr: str, port: str, MTU: int = 10000, tcp_no_delay: bool = False)

An easy temporary fix is done by editing the generated py file and changing the port value (8100) to a string.

daniestevez:
The port parameter of the socket_pdu block should be an str. Using an int
instead gives an error under GNU Radio 3.9 (and probably 3.10), since
pybind11 insists that the type must be correct. I think that under
GNU Radio 3.8 (with swig), an int parameter was converted silently to
an str, so we saw no problems.

This fixes #342

### DIFF
--- a/python/components/datasinks/CMakeLists.txt
+++ b/python/components/datasinks/CMakeLists.txt
@@ -49,4 +49,5 @@ include(GrTest)
 
 set(GR_TEST_TARGET_DEPS gnuradio-satellites)
 set(GR_TEST_PYTHON_DIRS ${CMAKE_BINARY_DIR} ${CMAKE_BINARY_DIR}/swig)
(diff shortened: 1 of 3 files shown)


daniestevez:
Besides fixing the bug, this also adds a QA test that tries to instance the kiss_server_sink.

PR #345, opened 2022-03-24
Bump to 3.10/3.11

bkerler:
This updates the module to gnuradio 3.10/3.11. Tested on both.


daniestevez:
Thanks! These files are used less frequently, so I hadn't noticed that they were grabbing the filter windows from `firdes` instead of `window`.

I see that some of the files, such as `examples/ber/lilacsat1_bpsk.grc` are indeed quite old, because they are XML flowgraphs, which date back to GNU Radio 3.7. The proper way to update these would be to open them with GNU Radio 3.10 and save them again. Would you like to do this in this PR?

bkerler:
Sure, can do :)

bkerler:
Fixes some more things and updated the examples. However I found one issue that I wasn't able to fix:
In examples/ccsds/Space_Packet_with_TimeStamp.grc the block variable_time_format_parameters cannot be parsed,
most likely because the grc/ccds/variable_time_format_parameters.block.yml is broken.

examples/ax25/bpsk.py
```diff
@@ -0,0 +1,325 @@
+#!/usr/bin/env python3
```
daniestevez, line 1: This file was not included before. I think it shouldn't be included in the repository, since users can generate it from the GRC file.

daniestevez (changes requested):

daniestevez:
> most likely because the grc/ccds/variable_time_format_parameters.block.yml is broken.

I agree that `variable_time_format_parameters.block.yml` is broken. This is something that was originally done for GNU Radio 3.7 and I converted it automatically to YAML with gr_modtool. I never really checked if it worked, but looking at the generated YAML, it seems gr_modtool didn't do a fine job.

The original XML file is at https://github.com/daniestevez/gr-satellites/blob/91234e539c835e7c6a3fed6ea370b42330c66751/grc/CCSDS/variable_time_format_parameters.xml in case you want to take a look and try to make the YAML by hand.

examples/ax25/bpsk.py
bkerler: Yep, sorry will remove it and add a .gitignore file as well to prevent similar issues

bkerler: Removed and added .gitignore.

bkerler:
> 

It took some time and even more patience, but it's finally working.

daniestevez:
Great! This looks good to me. Do you think this is ready to merge?

bkerler:
Yes, I've tested it on your workshop materials and it looks fine now.

PR #349, opened 2022-04-13
Delete DSTAR-One telemetry parser from cmake

argilo:
The corresponding GRC file was removed in a previous commit (18684a6b9f5a446db2735ddb59f090d8558ec7d6), which breaks installation.

Signed-off-by: Clayton Smith <argilo@gmail.com>


argilo:
I also noticed that the `dstar_one.grc` example references this block. What should be done about that?

https://github.com/daniestevez/gr-satellites/blob/31af3286a8245474eaa53856508342581025bd16/examples/satellites/nx_decoder/dstar_one.grc#L296-L308

PR #350, opened 2022-04-13
Modernize logging and remove all usage of Boost

argilo:
GNU Radio's logging framework now uses fmtlib instead of Boost.Format. GNU radio is also moving away from logging macros. I've updated logging here to match GNU Radio conventions.

I also switched a couple other instances of boost::format to standard C++.

Signed-off-by: Clayton Smith <argilo@gmail.com>

### DIFF
--- a/lib/crc_append_impl.cc
+++ b/lib/crc_append_impl.cc
@@ -82,7 +82,7 @@ void crc_append_impl::msg_handler(pmt::pmt_t pmt_msg)
 
     const auto size = msg.size();
     if (size <= d_header_bytes) {
-        GR_LOG_WARN(this->d_logger, "PDU too short; dropping");
+        this->d_logger->warn("PDU too short; dropping");
         return;
     }
 
--- a/lib/crc_check_impl.cc
+++ b/lib/crc_check_impl.cc
@@ -88,7 +88,7 @@ void crc_check_impl::msg_handler(pmt::pmt_t pmt_msg)
 
     const auto size = msg.size();
     if (size <= d_header_bytes + num_bytes) {
-        GR_LOG_WARN(this->d_logger, "PDU too short; dropping");
+        this->d_logger->warn("PDU too short; dropping");
         return;
     }
 
@@ -111,9 +111,9 @@ void crc_check_impl::msg_handler(pmt::pmt_t pmt_msg)
 
     const bool crc_ok = crc_computed == msg_crc;
     if (crc_ok) {
-        GR_LOG_INFO(this->d_logger, "CRC OK");
+        this->d_logger->info("CRC OK");
     } else {
-        GR_LOG_INFO(this->d_logger, "CRC fail");
+        this->d_logger->info("CRC fail");
     }
 
     const auto out_size = d_discard_crc ? size - num_bytes : size;
--- a/lib/decode_ra_code_impl.cc
+++ b/lib/decode_ra_code_impl.cc
@@ -79,7 +79,7 @@ void decode_ra_code_impl::msg_handler(pmt::pmt_t pmt_msg)
                 "message length: %ld, expected: %d\n",
                 (long)length,
                 ra_code_length * RA_BITCOUNT);
-        GR_LOG_ERROR(d_logger, "Invalid message length");
+        d_logger->error("Invalid message length");
         return;
     }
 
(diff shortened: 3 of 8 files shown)


argilo:
Note: I did not do any testing, I only verified that this builds against GNU Radio's `maint-3.10` and `main` branches.

daniestevez:
Great! I was thinking of doing this, but the thing holding me back was that I didn't know if this "modernized logging" was available in GNU Radio 3.9. Currently the main branch of gr-satellites targets both GNU Radio 3.9 and 3.10, and releases on that branch should work with both versions (Although a maint-3.9 branch could be created if there is enough need, but having a single branch for 3.9 and 3.10 simplifies maintenance).

The pipeline for 3.9 has passed, and I've seen this uses 3.9.5. So it seems that the modern logging also works with 3.9. Do you know if it also works with older 3.9 releases? I'll test with the 3.9.4 I have on my machine.

daniestevez (approved):

argilo:
As far as I know, the transition to fmtlib only happened in 3.10, so I'd be surprised if this works properly in 3.9. I'll try building locally to see how it goes.

daniestevez:
I'm testing it on 3.9.4 right now. It builds and runs, but it doesn't format the logging properly:
```
$ gr_satellites AO-73 --wavfile satellite-recordings/ao73.wav 
gr::debug :INFO: vmcircbuf_sysconfig - Using gr::vmcircbuf_sysv_shm_factory
gr::log :DEBUG: decode_rs0 - Reed-Solomon decode corrected {:d} bytes (interleaver path {:d})
gr::log :DEBUG: decode_rs0 - Reed-Solomon decode corrected {:d} bytes (interleaver path {:d})
-> Packet from 1k2 BPSK downlink
Frame type WO10
```

It seems it's not understanding the format string and simply ignoring the rest of the parameters.

daniestevez:
The other weird thing I found is [this compile warning](https://github.com/daniestevez/gr-satellites/runs/6012687612?check_suite_focus=true#step:4:150) about a potentially uninitialized variable, which I think didn't happen before. I checked the code and I think the compiler isn't right: `golay_res` will always be initialized at that point, because no one is changing the value of `d_use_golay` meanwhile. But I guess it doesn't hurt to initialize `golay_res` to 0 to suppress the warning.

argilo:
> I'm testing it on 3.9.4 right now. It builds and runs, but it doesn't format the logging properly
> It seems it's not understanding the format string and simply ignoring the rest of the parameters.

Ah, that's about what I expected, although I'm surprised the extra parameter don't cause a compile error.

I guess this change will have to wait until GNU Radio 3.9 support is no longer needed (or lives on a separate branch).

daniestevez:
Since apparently it's not so easy to get rid of boost::format in 3.9, and if I understood correctly https://github.com/gnuradio/gnuradio/pull/5645 makes gr-satellites current main not work any more with GNU Radio's main until this PR is merged, I think this maybe motivates spinning a maint-3.9 branch from main and then merging the PR to main. Any other alternatives or advice?

argilo:
> if I understood correctly https://github.com/gnuradio/gnuradio/pull/5645 makes gr-satellites current main not work any more

It compiles, but I haven't tested whether the logging works.

argilo:
It works fine. I guess `boost::format` is still accepted despite https://github.com/gnuradio/gnuradio/pull/5645.

argilo:
> I think this maybe motivates spinning a maint-3.9 branch from main and then merging the PR to main. Any other alternatives or advice?

It looks like could just wait until GNU Radio 3.9 support is no longer needed. Feel free to close off this pull request if you'd like to do that.

daniestevez:
Indeed I don't see any advantage to merging this now. But thanks so much for the effort. I'll close the PR, but when the time comes to modernize the logging, probably it's easier to start with this changeset to do most of the work.

Regarding dropping support for 3.9, I wonder if/when I'll be dropping support for 3.8, and don't see it coming soon, so probably 3.9 will be still for quite a long time.

daniestevez:
I have needed to split the gr-satellites code bases for GR 3.9 and 3.10 due to a totally unrelated issue (`vector_type` being moved from `gr::blocks::pdu` in gr-blocks to `gr::types` in gr-runtime, which makes it too difficult to share the same code base if `vector_type` is used in the API of a C++ block).

I've recovered the change set from this PR and put it in #359. I will merge it when I've checked that CI passes.

PR #356, opened 2022-05-23
Add support for CTIM 70 cm sw_stat beacons

aefyhrie:
Add support to gr-satellites for the Compact Total Irradiance Monitor (CTIM):

- Add a YAML definition for CTIM with temporary NORAD ID 99402
- Add a .py telemetry decoder for 70cm sw_stat beacon packets
- Make corresponding changes to CMakeLists.txt and init.py to include the new telemetry decoder in the build

### DIFF
--- a/python/satyaml/CTIM.yml
+++ b/python/satyaml/CTIM.yml
@@ -1,0 +1,13 @@
+name: CTIM
+norad: 99402
+data:
+  &tlm Telemetry:
+    telemetry: ctim_70cm
+transmitters:
+  9k6 FSK downlink:
+    frequency: 437.250e+6
+    modulation: FSK
+    baudrate: 9600
+    framing: AX.25 G3RUH
+    data:
+    - *tlm
--- a/python/telemetry/ctim.py
+++ b/python/telemetry/ctim.py
@@ -1,0 +1,365 @@
+# Copyright 2017, 2018, 2019, 2020 Daniel Estevez <daniel@destevez.net>
+# Copyright 2022 The Regents of the University of Colorado
+#
+# This file is part of gr-satellites
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+
+from .ax25 import Header
+from ..ccsds import space_packet as ccsds_space_packet
+from ..adapters import PolynomialAdapter
+from construct import Adapter, BitsInteger, BitStruct, Bytes, Container, \
+                Default, Enum, Flag, Float32l, Hex, If, Int8ub, Int8sl, \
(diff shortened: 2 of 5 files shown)


python/satyaml/CTIM.yml
```diff
@@ -0,0 +1,13 @@
+name: CTIM
+norad: 99402
+data:
+  &tlm Telemetry:
+    telemetry: ctim_70cm
```
daniestevez, line 5: The telemetry definition is called `ctim` rather than `ctim_70cm` in `python/telemetry/__init__.py` and `python/telemetry/ctim.py`. The names need to match. Maybe `ctim_70cm` is a better name in case support for the S-band telemetry gets eventually added (and also for consistency with `cute_70cm`)?

python/telemetry/ctim.py
```diff
@@ -0,0 +1,365 @@
...
+    'ax25_header' / Header,
+    'primary_header' / ccsds_space_packet.PrimaryHeader,
+    'secondary_header' / If(
+        lambda c: c.primary_header.secondary_header_flag,
+        SecondaryHeader
+    ),
+    'packet' / Switch(
+        lambda c: (c.primary_header.AP_ID),
+        {
+            (0x01): ctim_beacon
```
daniestevez, line 362: The parenthesis around  `0x01` are not necessary. Consider removing them.

python/telemetry/ctim.py
```diff
@@ -0,0 +1,365 @@
...
+
+ctim = Struct(
+    'ax25_header' / Header,
+    'primary_header' / ccsds_space_packet.PrimaryHeader,
+    'secondary_header' / If(
+        lambda c: c.primary_header.secondary_header_flag,
+        SecondaryHeader
+    ),
+    'packet' / Switch(
+        lambda c: (c.primary_header.AP_ID),
```
daniestevez, line 360: The parenthesis around  `c.primary_header.AP_ID` are not necessary. Consider removing them.

daniestevez (changes requested):
Thanks for your work! I think the code is quite good. I added a couple of comments about style that you may want to address (specially the one about `ctim` vs. `ctim_70cm`, since I think the `.yml` file won't work currently in its current state).

aefyhrie:
Hey @daniestevez , I made the requested changes -- good call with updating the naming convention to include _70cm to make it compatible with the yaml file and to leave room for the S-band downlinks if we choose to include those.

Please let me know if there is anything else I can do

daniestevez (approved):
Thank you! I'll now squash and merge this.

aefyhrie:
Thanks!

PR #366, opened 2022-06-13
Add example flowgraph for gqrx/UDP stream input

phl0:
This adds an example flowgraph that can be used with gqrx and an UDP stream of I/Q as source.


daniestevez:
Thanks! I've merged this manually in 65daf3f97a674669eaabfd688a863f32b72cd02d.

I did the manual merge because one of the commits was modifying the satellite-recordings submodule (this can happen inadvertently sometimes).

phl0:
Never mind. Tnx fer merge. Must have been a mistake. Will take more care next time.

PR #368, opened 2022-06-25
Create SUCHAI-2.yml

janvgils:
Based on https://twitter.com/EU1SAT/status/1540840856424484866 we can now determine that SUCHAI-2 is object 52192


daniestevez:
Thanks!

PR #371, opened 2022-07-01
Create RANDEV.yml

janvgils:
Newly launched cubesat - http://ascl.kaist.ac.kr/


daniestevez:
Thanks for keeping up with these launches.

PR #375, opened 2022-07-03
Update PicSat.yml

janvgils:
Add the PICSAT SIDS server so decoded data is forwarded to this server, one remark, the user needs to register and enable SIDS on the PICSAT website https://picsat.obspm.fr/

### DIFF
--- a/python/satyaml/PicSat.yml
+++ b/python/satyaml/PicSat.yml
@@ -1,5 +1,7 @@
 name: PicSat
-norad: 43132 
+norad: 43132
+telemetry_servers:
+  - SIDS https://picsat.obspm.fr/sids/reportframe?
 data:
   &tlm Telemetry:
     telemetry: picsat


daniestevez:
Thank you! Merging.

Do you know if to make this work is it sufficient to register an account in the PicSat web with your callsign, then enable SIDS on the PicSat web, and finally, ensure that you use the same callsign when submitting from gr-satellites using SIDS (this is in the `[Groundstation]` section in `~/.gr_satellites/config.ini`). gr-satellites doesn't send any authentication to SIDS servers, so if this has worked for you, I guess that they're only checking the callsign against their list of registered users.

janvgils:
Hi Daniel,

The call registered in ${HOME}/.gr_satellites/config.ini needs to be the same as registered on their website.
You also need to enable SIDS for your account on their website.

So yes, they use the source for authorization.

https://picsat.obspm.fr/sids/reportframe?noradID=100&source=F4HZG&timestamp=2018-01-12T09:00:00Z&frame=FE DC BA 98 76 54 32 10&locator=latlong&longitude=5E&latitude=45N

Jan

PR #380, opened 2022-07-13
Create MTCUBE-2.yml

janvgils:
Newly launched Satellite ni MEO orbit https://csum.umontpellier.fr/


daniestevez:
I've just merged this manually including the SIDS server.

Issue #159, opened 2020-09-06
Missing option, --f_offset not recognised 


janvgils:
With the lasted build and trying to decode a BPSK observation I am running in the following error:

`gr_satellites satellite: error: unrecognized arguments: --f_offset 12e3`

I am trying this with the following cli command:

`gr_satellites 46277 --wavfile UPMSat-2_satnogs_2798315_2020-09-06T11-09-46.wav --samp_rate 48e3 --f_offset 12e3 --hexdump`

```
gr_satellites --version 46277
gr_satellites v3.4.0-git
Copyright (C) 2020 Daniel Estevez
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```
Observation used: https://network.satnogs.org/observations/2798315/

Is the support for freq offsets dropped in the current releases?

daniestevez:
UPMSat-2 uses FSK, not BPSK, hence the lack of the `--f_offset` option.

Now some rationale about this. With FSK satellites, 99% of the time we're using recordings of FM-demodulated data (and I mean the whole community, not just gr-satellites). This is because traditionally people have used conventional radios in FM mode to record these, and now they're using SDR software in FM mode. In fact, this is actually not so good. The fact that FM mode is done before the decoder precludes other better methods of decoding. It would be better to record FSK signals in "wide-SSB" mode, so that they fit into the 24kHz passband of a 48ksps real signal, or in IQ mode, but people very rarely do this.

With this in mind, gr-satellites works as described in [FSK demodulation and IQ input](https://gr-satellites.readthedocs.io/en/latest/command_line.html#fsk-demodulation-and-iq-input). Basically, if you don't use the `--iq` option, it expects an audio signal which is already FM-demodulated. If you use `--iq`, it expects an RF signal which still has FM. This may be counter-intuitive and somewhat strange, because changing to IQ mode changes the demodulation method, but I figured that this is what would work with most of the recordings out there, since most non-IQ recordings are already FM-demodulated and all IQ recordings have FM.

Now we come to UPMSat-2. This satellite uses 1k2 FSK, which is a bit uncommon, given its low baudrate. In my opinion, this should be recorded or demodulated in FM mode, just as we do with other FSK signals. As mentioned above, it is not the best approach, but it is consistent with what we do with other satellites.

[...]

PR #392, opened 2022-08-06
Add YML file for AzaadiSAT

K4KDR:
ISRO launch scheduled for 7-Aug-2022

Details:  https://community.libre.space/t/sslv-d1-eos2-azaadisat-mission/9478/

Using SX1268 chip, multiple modes (LoRa, SSTV, CW, FSK).      Using FSK test audio file provided by the Project Team, was able to determine non-standard FSK (really AFSK, I think) tones which are reflected in this YML file.   

Sample audio file & decode text available at ->  https://www.qsl.net/k/k4kdr//audio/azaadisat/

Only tested in 'maint-3.8' branch; I assume this would be compatible with other branches but only tested in 'maint-3.8'

### DIFF
--- a/python/satyaml/AzaadiSAT.yml
+++ b/python/satyaml/AzaadiSAT.yml
@@ -1,0 +1,15 @@
+name: AzaadiSAT
+norad: 99999
+data:
+  &tlm Telemetry:
+    unknown
+transmitters:
+  1k2 AFSK downlink:
+    frequency: 437.400e+6
+    modulation: AFSK
+    baudrate: 1200
+    af_carrier: 3625
+    deviation: -625
+    framing: AX.25
+    data:
+    - *tlm


daniestevez:
Hi Scott,

I don't think that the SX1268 can produce AFSK. My interpretation is that the WAV file you have is FSK, but has been recorded "in USB mode" rather than in "FM mode".

The correct approach would be to use `mode: FSK` in the SatYAML file, and omit the `af_carrier` and `deviation` fields. However, currently there is no way of decoding such a WAV file through the `gr_satellites` command line (unless you trick it by telling that it's AFSK). This is somewhat tricky (see #159). A good way to process this file is to use a Quadrature Demod block in GNU Radio Companion (after shifting the signal to baseband) and then the gr-satellites Satellite Decoder block.

K4KDR:
Fascinating!  I was suspicious & curious about the sample audio file that was provided... I was not clear on how it was produced.

For these reasons I was reluctant to post info regarding Direwolf settings & to propose a YML file here, but with the launch imminent, I wanted to share what info I did have with the community.

Thanks so much for providing some badly needed clarity to this situation!  You can feel free to modify or delete this request in whatever manner is most useful to gr-satellites.

PR #397, opened 2022-08-16
Add Tsiolkovsky-Ryazan 2

janvgils:
One the other RADIOSKAF launched satellites.


daniestevez:
Hi Jan,

The parameters for the baudrate and deviation seem somewhat strange. Are these published somewhere or did you measure them?

At least I noticed that 1135/565 is close to 2. A deviation of 1/2 of the baudrate is common to have orthogonal tones (MSK would be 1/4 of the baudrate). So if the baudrate really is 1135, then I guess the deviation should be 567.5 Hz (exactly 1/2 of the baudrate).

janvgils:
Hi Daniel,

Yes, you are right, these values are very strange, I used MixW3 to measure the signals.
Let me record some examples and share them.


daniestevez:
Thanks. I think it will be worth to take a close look to some recordings to ensure that the parameters are correct.

daniestevez:
I have taken a look at a recording of these satellites ([this SatNOGS observation](https://network.satnogs.org/observations/6398066/)) to try to measure the baudrate and deviation. The baudrate seems closer to 1136 baud than to 1135, and I couldn't get a precise enough measurement of the deviation because the transmitter seems a bit wobbly (specially for the low tone, see figure below).

I think that the parameters in the SatYAML file are more or less right, and that measuring more accurately is not easy, so I'll cherry-pick 17ae8b5ba78095a78ebaac7158406e98b7f9059b into main (the [remaining commits](https://github.com/daniestevez/gr-satellites/pull/397/commits), in this pull request don't really belong here, and I think they're all already in main).

inst_freq

janvgils:
Good morning Daniel,

I have used following observation: https://network.satnogs.org/observations/6378822/
It isn't the best but there is one frame in there that has enough SNR for decoding.

Replay and decoding with MixW3:

MixW sample rate detection shows it as 1129 baud.
[cid:6ac59572-9316-4bf0-96f1-31876f2c66a1]

cat TSIOLKOVSKY-RYAZAN-2.yml

name: TSIOLKOVSKY-RYAZAN-2
alternative_names:
  - RS12S
norad: 53313
data:
  &tlm0 AX25 telemetry:
    telemetry: ax25
transmitters:
  1k2 AFSK downlink:
    frequency: 437.0125e+6
    modulation: AFSK
    baudrate: 1129
    af_carrier: 1700
    deviation: 565
    framing: AX.25
    data:
    - *tlm0

Decoding the same frame with gr-satellites:
I changed the deviation to 565.

​gr_satellites TSIOLKOVSKY-RYAZAN-2.yml --samp_rate 48e3 --audio
-> Packet from 1k2 AFSK downlink
Container:
    header = Container:
        addresses = ListContainer:
            Container:
                callsign = u'ALL' (total 3)
                ssid = Container:
                    ch = True
                    ssid = 0
                    extension = False
            Container:
                callsign = u'RS12S' (total 5)
                ssid = Container:
                    ch = False
                    ssid = 0
                    extension = True
        control = 0x03
        pid = 0xF0
    info = b'KETs #2 S=780 U1=4.36 U2=4.27 I1=0.0 T1=9.50 Mx=22.650711 My=-28.033058 Mz=32.966877 Vx=16.442749 Vy=119.206108 Vz=15.351145 Time=-1657640171 R1=0 R2=0' (total 151)

Jan PE0SAT

daniestevez:
Hi, I'm not sure how accurate this MixW measurement is supposed to be. In any case, the difference between 1135 (which is what we currently have) and 1129 is relatively small, so I don't think it's worth changing anything.

janvgils:
Hi Daniel,

Thanks for the reply, that is fine with me.

PR #408, opened 2022-10-12
Remove unreachable code list_satellites()

kng:
list_satellites() handled by class ListSatellites()

Signed-off-by: Daniel Ekman <knegge@gmail.com>

### DIFF
--- a/apps/gr_satellites
+++ b/apps/gr_satellites
@@ -176,9 +176,6 @@ class gr_satellites_top_block(gr.top_block):
                       file=sys.stderr)
         else:
             options = parser.parse_args(satcfg_args + sys.argv[2:])
-        if options.list_satellites:
-            list_satellites()
-            sys.exit(0)
         check_options(options, parser)
 
         self.options = options
@@ -307,8 +304,6 @@ def main():
     if len(sys.argv) >= 2 and sys.argv[1] in ['--version',
                                               '--list_satellites']:
         options = parser.parse_args()
-        if options.list_satellites:
-            list_satellites()
         sys.exit(0)
 
     if len(sys.argv) <= 1 or sys.argv[1][0] == '-':


daniestevez:
What is the reasoning behind this PR? The code is not unreachable. It is used to implement the `--list_satellites` option.

kng:
There isn't any function called list_satellites() anymore ?
Afaict it's handled directly by [the action](https://github.com/daniestevez/gr-satellites/blob/main/apps/gr_satellites#L64) and the corresponding [class](https://github.com/daniestevez/gr-satellites/blob/main/apps/gr_satellites#L34).

daniestevez:
You're right.

Incidentally, both the `ListSatellites` action and the calls to the (inexistent) function `list_satellites()` were introduced in the same commit 2bc9a8fbc6b0969f78ce3488f525ee58ebd9ae5b, so I don't know what I was thinking when I did this. I guess the calls to `list_satellites()` were leftovers from a preliminary attempt at implementing this.

apps/gr_satellites
```diff
@@ -307,8 +304,6 @@ def main():
     if len(sys.argv) >= 2 and sys.argv[1] in ['--version',
                                               '--list_satellites']:
```
daniestevez, line 305: We should also remove the `'--list_satellites'` option from here, since it's not useful at all (as you pointed out, the `--list_satellites` option is immediately handled by the action).
The code here should read `if len(sys.argv) >= 2 and sys.argv[1] == '--version'`.

daniestevez (changes requested):

apps/gr_satellites
kng: That results in useage being shown, the actual parse_args() needs to be reached ?

daniestevez: I see. Indeed. I agree this looks good, so let's merge it.

daniestevez (approved):

PR #409, opened 2022-10-16
Update TEVEL-1.yml

janvgils:
Wrong NoradID and changed it. 

https://celestrak.org/NORAD/elements/gp.php?CATNR=51013


daniestevez:
Thanks for spotting this!

PR #421, opened 2022-11-29
Fix baudrate for SWSU satellites

sv1:


sv1:
Hello  Daniel,

unfortunately I can not find any official documentation but this is what I have used and successfully decoded telemetry with direwolf and gr_satellites.

daniestevez:
Does it fail to decode with the baudrate we currently have in the repository?

sv1:
Yes! that is why I started looking into this and made that change!

daniestevez:
Can you share the recordings you used? It's interesting because earlier when we attempted to measure the baudrates of these satellites it turned out that the baudrates were around 1130-1135, which is well away from 1145. It could be that their baudrates are not so stable and keep changing.

sv1:
I do not have recordings, only real-time decoding!

daniestevez:
Then maybe it's useful to record the next time you try to receive.

daniestevez:
Closing, since this has long been inactive and I'm not keen on changing the baudrates without some kind of precise measurement.

PR #436, opened 2023-02-10
Create AZAADISAT-2.yml

janvgils:
New launch https://twitter.com/SpaceKidzIndia/status/1623643412250652672

### DIFF
--- a/python/satyaml/AZAADISAT-2.yml
+++ b/python/satyaml/AZAADISAT-2.yml
@@ -1,0 +1,13 @@
+name: AZAADISAT-2
+norad: 99198
+data:
+  &tlm Telemetry:
+    telemetry: ax25
+transmitters:
+  1k2 FSK downlink:
+    frequency: 437.275e+6
+    modulation: FSK
+    baudrate: 1200
+    framing: AX.25
+    data:
+    - *tlm


daniestevez:
Hi, thanks for this.

I was reviewing [this document](https://github.com/Space-Kidz-India/AzaadiSAT-2_RF_Parameters/blob/main/AzaadiSAT-2%20-%20RF%20Parameters.pdf) and thinking that it doesn't say clearly that it uses FSK AX.25. It could either be AFSK or FSK, although "Frequency shift: 425 Hz" perhaps points more to FSK, since the standard AFSK tones do not use this shift. I'm also worried by "Stop bits: 1", which is something that belongs to UART and has nothing to do with HDLC, which is what AX.25 does. Since it seems they've used a LoRa chip, if they haven't implemented things properly, the outcome won't be AX.25 compatible.

I thought of comparing with AzaadiSat-1, and found that we already discussed similar things in #392. There we had a sample recording and apparently we were able to decode it as FSK AX.25.

Therefore, I'm going to assume that AzaadiSat-2 is equal to AzaadiSat-1 in this respect and merge this. It will be good to test the decoder with some recordings.

daniestevez:
Just seen some decodes in Twitter using FSK AX.25: https://twitter.com/scott23192/status/1623971689359978498

K4KDR:
Downloaded **'AZAADISAT-2.yml'** and can confirm that it decodes properly from an FM demod audio file that I recorded from I/Q replay after one of the passes here last night.


gr-satellites-YAML-decode

PR #440, opened 2023-02-21
Update KUZBASS-300.yml

janvgils:
Add 19k2 to the yml file, as can be seen in this observation https://network.satnogs.org/observations/7186911/ the satellite is also capable of 19200 baud.


daniestevez:
Looks good. Merging.

Issue #295, opened 2021-08-29
CSP header swap for AX100 transceiver


daniestevez:
Following some messaging with Mark Jessop, I think it's probably time to eliminate the [swap of the CSP header](https://github.com/daniestevez/gr-satellites/blob/717c7540e266957aba43f9665da3accc2fe5bad9/lib/ax100_decode_impl.cc#L80) that exists for historical reasons. There is a lot of background about why this is in place [in this comment in the librespace forums](https://community.libre.space/t/standardization-of-frames-formats-for-satnogs-db/7057/2?u=ea4gpz).

It would be good to adopt a consistent solution between all the applications that submit frames to SatNOGS DB, including gr-satellites, gr-satnogs (which does not swap the header), and the UZ7HO soundmodem (which I'm not sure what it currently does).

darksidelemm:
According to Patrick Dohmen, the UZ7HO modem doesn't do any header swapping.

daniestevez:
Just say Patrick's message. So apparently after UZ7HO dropped the swap earlier this year, gr-satellites is the last application that is doing the swap. I'll remove the swap now.

PR #446, opened 2023-04-05
Changed CSP packet id to network endian

Rakagami:
In libcsp the Header ID is encoded as network/big-endian for csp 1 and csp 2.0:

Refer to:
- https://github.com/libcsp/libcsp/blob/6f81c3da04c051c35ed6c28612a4037c031bb1f3/src/csp_id.c#L42-L59
- https://github.com/libcsp/libcsp/blob/6f81c3da04c051c35ed6c28612a4037c031bb1f3/src/csp_id.c#L123-L141

### DIFF
--- a/python/csp_header.py
+++ b/python/csp_header.py
@@ -15,7 +15,7 @@ class CSP(object):
     def __init__(self, csp_packet):
         if len(csp_packet) < 4:
             raise ValueError('Malformed CSP packet (too short)')
-        csp = struct.unpack('<I', csp_packet[0:4])[0]
+        csp = struct.unpack('!I', csp_packet[0:4])[0]
         self.priority = (csp >> 30) & 0x3
         self.source = (csp >> 25) & 0x1f
         self.destination = (csp >> 20) & 0x1f


daniestevez:
The file `python/csp_header.py` is only used in some legacy code. We now have [`python/telemetry/csp.py`](https://github.com/daniestevez/gr-satellites/blob/main/python/telemetry/csp.py), which uses construct and is used mostly everywhere.

I've checked where `csp_header.py` is still used, and saw the following:
* `check_crc.py`. This block is deprecated and will disappear.
* `print_header.py`. This block is deprecated and will disappear.
* `swap_crc.py`. This block isn't marked as deprecated, but it isn't used anywhere in gr-satellites.

The endianness of in `csp.py` was changed at some point because of #295. Your PR now changes `csp_header.py` to match the current state of `csp.py`, which seems good. Chances for breaking something else as a result of this change are slim, because `csp_header.py` isn't really used anywhere. Therefore, I'm happy to merge this PR, even though `csp_header.py` should be removed at some point. If you have code that is using `csp_header.py`, it probably makes sense to update it to use `csp.py`.

Rakagami:
We are using `print_header.py` in our project which depended on `csp_header.py`. We will probably update to `csp.py` then if `csp_header.py` will be deleted anyway

daniestevez:
It's possible to use the [Telemetry Parser block](https://gr-satellites.readthedocs.io/en/latest/components.html#telemetry-parser) with a telemetry definition of `csp` to replace `print_header.py`.

And now that I check, I see that the import for `cspv2` is missing in `__init__.py`, so I'll go ahead and add this.

PR #462, opened 2023-06-09
usp_ax25_crop grc block fix

baskiton:
change pls to correct ax25


daniestevez:
Thanks for spotting this bug. It looks good. Merging.

PR #477, opened 2023-07-05
BME WebSocket telmetry submitter (maint-3.8 branch)

daniestevez:
This adds support for sending telemetry to the BME WebSocket server, which is used by MRC-100.

The implementation uses the Python websocket-client package (Debian /Ubuntu package python3-websocket), which has been added as a dependency in the installation instructions in the documentation and in the debian package control file. The build-ubuntu CI action has been updated to include python3-websocket in the Docker image.

The older BME HTTP telemetry server has been marked as deprecated, since it is no longer operational.

### DIFF
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -13,4 +13,4 @@ jobs:
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v3
-      - uses: daniestevez/gr-satellites-ci-action@v0.1.3
+      - uses: daniestevez/gr-satellites-ci-action@v0.1.4
--- a/debian/control
+++ b/debian/control
@@ -13,7 +13,8 @@ Build-Depends: cmake,
                python3-six,
                swig,
                python3-construct,
-               python3-requests
+               python3-requests,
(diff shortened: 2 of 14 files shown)


K4KDR:
Thank you for adding this feature!

I see the following at https://gr-satellites.readthedocs.io/en/latest/command_line.html?highlight=bme#telemetry-submission

`The BME server (WebSocket) does not require any registration or additional configuration.`

... and I also see that **~/gr-satellites/python/satyaml/MRC-100.yml**  includes:

```
telemetry_servers:
  - BMEWS
```

My question is, does that entry in the .yml file cause the telemetry to be submitted automatically?

Or, do we also need to incorporate a **BME Telemetry Forwarder (WebSocket)** block in our flowgraphs?

Thanks!

daniestevez:
Telemetry is submitted automatically when using the `gr_satellites` command line tool (as long as you have telemetry submission enabled in `.gr_satellites/config.ini`). If you're using your own flowgraph, you need to include either the `BME Telemetry Forwarder (WebSocket)` block or the `Telemetry Submit` (and set it to "BME Telemetry Server (WebSocket)") in the flowgraph.

K4KDR:
Wonderful!   I wasn't clear on that.

Thanks so much!

PR #504, opened 2023-07-21
Update AzaadiSAT.yml

janvgils:
Changed the Norad-ID to the SatNOGS temporary and also added the SatNOGS id as an alternative name. This could open up the use of the satnogs id's when decoding the satellites.

The only question is, are alternative names always available in the gr-satellite functions ?

### DIFF
--- a/python/satyaml/AzaadiSAT.yml
+++ b/python/satyaml/AzaadiSAT.yml
@@ -1,5 +1,7 @@
 name: AzaadiSAT
-norad: 99999
+alternative_names:
+  - EPDM-0969-2670-9539-4005
+norad: 99355
 data:
   &tlm Telemetry:
     telemetry: ax25
--- a/python/satyaml/AzaadiSAT.yml
+++ b/python/satyaml/AzaadiSAT.yml
@@ -1,6 +1,4 @@
 name: AzaadiSAT
-alternative_names:
-  - EPDM-0969-2670-9539-4005
 norad: 99355
 data:
   &tlm Telemetry:


daniestevez:
Alterative names are used to match the satellite by name. Any time that you can use the main satellite name, you can substitute by an alternative name.

Before setting ourselves to adding SatNOGS UUIDs as alternative names, I'd rather question the usefulness of it. It's a lot of work to add UUIDs retroactively for all the existing SatYAMLs, unless we do this with a script (I don't know if the SatNOGS DB API would be queryable for this). The main reason why we have NORAD IDs in the SatYAML is because they're used when submitting telemetry to SatNOGS or another SiDS server. If this wasn't the case, maybe we wouldn't have NORAD IDs in the SatYAML. We don't currently have a way to submit telemetry from gr-satellites using UUIDs instead (we never implemented something like SatNOGS artifacts). Is there an application where UUIDs might be helpful (maybe when using gr-satellites integrated into gr-satnogs)? If so, then maybe I'd rather have a specific mandatory field for the UUID instead of using the alternative name, and would look into ways of filling this automatically at least for existing files.

janvgils:
Good evening, Daniel,


I agree, if UUID's will be the way in the future we will need to find another option then adding it as an Alternative name.

For now, I was just looking into what is possible if I have the UUID available in some way when using gr_satellites in combination with satnogs.


For now, the only thing unique is the Norad ID and that will always be a pain at the beginning of a mission.
Could the SatNOGS UUID be an alternative?


Jan

kng:
> (I don't know if the SatNOGS DB API would be queryable for this)

https://db.satnogs.org/api/satellites/?norad_cat_id=42017

daniestevez:
> Could the SatNOGS UUID be an alternative?

I'm not sure, since I don't know if we can submit telemetry to SatNOGS DB using UUIDs. I agree that using temporary NORAD IDs at the beginning of a mission is inconvenient. It would be good to point some SatNOGS people to this thread to get their views on 3rd party telemetry clients, as well as the usual stakeholders for other clients (DK3WN...).

daniestevez:
> > (I don't know if the SatNOGS DB API would be queryable for this)
> 
> https://db.satnogs.org/api/satellites/?norad_cat_id=42017

Thanks! That makes it possible to easily automate adding a UUID to a SatYAML file by querying the DB API.

kng:
More questions arise; should name be considered primary ? should satcfg use norad or name ? what more mechanisms is affected here ?

janvgils:
As Daniel (sa2kng) already mentioned in another conversation, this is a can of worms.

We will all need to agree on a single point of truth, for everyone using the SatNOGS db to forward the decoded telemetry.
So Mike DK3WN also needs to be part of this and all other forwarding telemetry to the db.

I know that Mike already tried to use the SatNOGS API in combination with his getkiss+ he queried the API but regular got a query time out.

But for now, will we remove the UUID from the yml files an if so how can this be done working with git ?

daniestevez:
> But for now, will we remove the UUID from the yml files an if so how can this be done working with git ?

Removing the UUID from the YML files for now sounds good.

janvgils (commented):
Changed the Pull request

PR #505, opened 2023-07-21
Update FOSSASAT-1B.yml

janvgils:
Changed to SatNOGS temporary Norad and adding SatNOGS id

### DIFF
--- a/python/satyaml/FOSSASAT-1B.yml
+++ b/python/satyaml/FOSSASAT-1B.yml
@@ -1,6 +1,7 @@
 name: FOSSASAT-1B
 alternative_names:
-norad: 99999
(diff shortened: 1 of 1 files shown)


python/satyaml/FOSSASAT-1B.yml
```diff
@@ -1,6 +1,7 @@
 name: FOSSASAT-1B
 alternative_names:
-norad: 99999
+  - SHUV-9360-7405-4358-2470
```
daniestevez, line 3: Same remark as in https://github.com/daniestevez/gr-satellites/pull/504#issuecomment-1646084372

Let's wait to decide how we want to handle SatNOGS UUIDs before merging this.

PR #508, opened 2023-07-24
refactor snet support

Gioyik:
As part of a university project, I got access to more information related to SNET satellites, this adds and introduces changes in telemetry to get closer to the specification of SNET.

### DIFF
--- a/grc/satellites_snet_deframer.block.yml
+++ b/grc/satellites_snet_deframer.block.yml
@@ -9,7 +9,7 @@ parameters:
     options: ['True', 'False']
     option_labels: ['Yes', 'No']
 -   id: buggy_crc
-    label: Verbose
+    label: Buggy CRC
     dtype: bool
     default: True
 
--- a/python/telemetry/snet.py
+++ b/python/telemetry/snet.py
@@ -11,13 +11,14 @@
 import datetime
 
 from construct import *
-import construct
 
 from ..adapters import LinearAdapter
 
 # See
-# https://www.raumfahrttechnik.tu-berlin.de/fileadmin/fg169/
-# amateur-radio/TUBiX10-COM.pdf
+# 1. https://www.static.tu.berlin/fileadmin/www/10002275/Amateur_Radio/
+#    TUBiX10-COM.pdf
+# 2. https://www.static.tu.berlin/fileadmin/www/10002275/Amateur_Radio/
+#    QldqXP-TUBIX10_3800_TN03_TM_Frame_01.pdf
 # for documentation
 
 LTUFrameHeader = BitStruct(
@@ -56,8 +57,7 @@ class TimeAdapter(Adapter):
 
 TimeStamp = TimeAdapter(BitsInteger(32, swapped=True))
 
-
-SNETFrameHeaderExtension = BitStruct(
+PDUFrameHeaderExtension = BitStruct(
     'VersNo' / BitsInteger(2),
     'DFCID' / BitsInteger(2),
     'ExtensionRFU' / BitsInteger(4),
@@ -71,7 +71,7 @@ SNETFrameHeaderExtension = BitStruct(
     'SeqNo' / BitsInteger(14)
     )
 
-SNETFrameHeader = BitStruct(
+PDUFrameHeader = BitStruct(
     Const(0b111100110101000000, BitsInteger(18)),
     'CRC' / BitsInteger(14),
     'FCIDMajor' / BitsInteger(6),
@@ -84,8 +84,7 @@ SNETFrameHeader = BitStruct(
     'TimeTagged' / Flag,
     'DataLength' / BitsInteger(10),
     'TimeTag' / If(lambda c: c.TimeTagged, TimeStamp),
-    'Extension' / If(lambda c: c.Extended, SNETFrameHeaderExtension)
-    )
+    'Extension' / If(lambda c: c.Extended, PDUFrameHeaderExtension))
 
 Battery = Struct(
     'V_BAT' / LinearAdapter(2, Int16sl),
@@ -168,10 +167,54 @@ ADCSTelemetry = Struct(
     'TargetData_ControlIsActive' / Int8ul  # flag, really
     )
 
+# Mailbox in SNET satellites
+# There are three main fields defining the Mailbox inside SNET satellites:
+#
+# Name        ByteLength  ParamType   TargetType    Interpretation
+# SysTime     4           uint32_t    float         scale: 0,001 unit: s
+# MemDataType 1           uint8_t     uint8_t       values: 0, 1, 2, 3, 4, 5
+# MemData32   32          uint8_t     Array         format: Array
+#
+# Notes:
+# * `MemDataType` refers to the slot of memory in the satellite where
+# the information is stored, which slot of memory.
+# * `MemData32` is an Array of bytes, where each byte corresponds to an ascii
+# character code. The transmission of words is done by Big-Endian order.
+
+
+class MemData32Adapter(Adapter):
+    def _encode(self, obj, context, path=None):
+        return bytes([x << 1 for x in bytes(
+            (obj.upper() + ' '*8)[:8], encoding='ascii')])
+
+    def _decode(self, obj, context, path=None):
+        # do we need to do right shift here?
+        return str(bytes([x for x in bytes(
+            (obj.upper() + ' '*8)[:8], encoding='ascii')]),
+                   encoding='ascii').strip()
+
+
+MemData32 = MemData32Adapter(Bytes(32))
+
+MailboxReceiveTelemetry = Struct(
+    'SysTime' / BitsInteger(32),
+    'MemDataType' / BitsInteger(8),
+    'MemData32' / MemData32
+    )
+
+MailboxSendTelemetry = Struct(
+    'MemDataType' / BitsInteger(8),
+    'MemData32' / MemData32
+    )
 
 snet = Struct(
-    'header' / SNETFrameHeader,
-    'telemetry' / Switch(lambda c: (c.header.FCIDMajor, c.header.FCIDSub), {
+    'header' / PDUFrameHeader,
+    'data' / Switch(lambda c: (c.FCIDMajor, c.FCIDSub), {
         (0, 0): ADCSTelemetry,
         (9, 0): EPSTelemetry,
-    }, default=Pass))
+        (54, 17): MailboxReceiveTelemetry,  # COM0
+        (54, 117): MailboxReceiveTelemetry,  # COM1
+        (56, 34): MailboxSendTelemetry,  # COM0
+        (56, 134): MailboxSendTelemetry,  # COM1
+    },
+      default=Pass))
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *~
 apps/*.py
 __pycache__
+.DS_Store


Gioyik:
I am not sure why the tests are failing, as it seems, they are not related to the changes I made.

daniestevez:
I think the tests are failing because the `satellites` Python module cannot be imported due to the changes you made. The error message in the tests is not very helpful, though, but you can check what `import satellites` says on your machine.

python/snet_deframer.py
```diff
@@ -165,9 +166,22 @@ def handle_msg(self, msg_pmt):
...
+        pdu_tags = pmt.dict_add(
+            pdu_tags, pmt.intern('FCIDSub'), pmt.to_pmt(pdu_parsed.FCIDSub))
+        pdu_tags = pmt.dict_add(
+            pdu_tags, pmt.intern('FCIDMajor'),
+            pmt.to_pmt(pdu_parsed.FCIDMajor))
+
         self.message_port_pub(
             pmt.intern('out'),
-            pmt.cons(pdu_tags, pmt.init_u8vector(len(pdu), pdu)))
+            pdu_tags)
```
daniestevez, line 187: Deframers in gr-satellites should output the frame as a GNU Radio PDU, which means a `pmt.cons()` of some metadata (not obtainable from the frame contents) and the frame bytes as a PMT u8vector). Any parsing of the frame into fields should be done in a telemetry parser. Sometimes it's difficult to say what a "frame" exactly should be (should it include the CRC?, should it include the length field if any?, etc.), but for S-NET I think that `pdu` is the frame. Therefore, the original code seems fine to me.

python/telemetry/snet.py
```diff
@@ -168,10 +100,91 @@ def _decode(self, obj, context, path=None):
...
+    'Arq' / Flag,
+    'PduTypeId' / Flag,
+    'BchRq' / Flag,
+    'Hailing' / Flag,
+    'UdFl1' / Flag,
+    'PduLength' / BitsInteger(10),
+    'CRC13' / BitsInteger(13),
+    'CRC5' / BitsInteger(5),
+    Padding(2)
+    )
```
daniestevez, line 122: Please do not needlessly move the location of this (and other elements) in the source code file. If it's move, it's difficult to tell if there are any changes by looking at the diff.

python/telemetry/snet.py
```diff
@@ -168,10 +100,91 @@ def _decode(self, obj, context, path=None):
...
+            milliseconds=d.seconds * 1000 + d.microseconds / 1000)
+
+    def _decode(self, obj, context, path=None):
+        return (datetime.datetime(2000, 1, 1)
+                + datetime.timedelta(seconds=float(obj)/2.0))
+
+
+TimeStamp = TimeAdapter(BitsInteger(32, swapped=True))
+
+PDUFrameHeaderExtension = BitStruct(
```
daniestevez, line 152: Is there a reason for renaming `SNETFrameHeaderExtension` to `PDUFrameHeaderExtension` and `SNETFrameHeader` to `PDUFrameHeader`?

python/telemetry/snet.py
```diff
@@ -168,10 +100,91 @@ def _decode(self, obj, context, path=None):
...
         (0, 0): ADCSTelemetry,
         (9, 0): EPSTelemetry,
-    }, default=Pass))
+        (54, 17): MailboxTelemetry,  # COM0
+        (54, 117): MailboxTelemetry  # COM1
+    },
+      # why parse instead of PASS? it would be good to have a repesentation
+      # in case the wave file contains a new pair of FCIDMajor and FCIDSub not
+      # supported. It creates a bit of lag, but I guess is worth it instead of
+      # losing the UserData.
```
daniestevez, line 189: The goal of a telemetry parser is to produce human readable output. All the possible telemetry formats that the satellite can transmit are supposed to be implemented in this file. For other satellites, often this is not possible due to lack of documentation or development effort, in which case sometimes a greedy bytes or similar is used for the payload of the frame.

Data should not be lost if a frame cannot be parsed, since the original data can be written to a KISS file.

python/telemetry/snet.py
```diff
@@ -168,10 +100,91 @@ def _decode(self, obj, context, path=None):
...
+      default=GreedyString("utf-8")))
```
daniestevez, line 190: Would this cause problems if the binary data is not valid as a UTF-8 string?

python/telemetry/snet.py
```diff
@@ -168,10 +100,91 @@ def _decode(self, obj, context, path=None):
...
+# To send mailbox messages
+# FCIDMajor = 56  (COM0)
+# FCIDSub = 34 || 134  (COM1)
+
+MailboxTelemetry = Struct(
+    'SysTime' / BitsInteger(4),
+    'MemDataType' / BitsInteger(1),
+    'MemData32' / Hex(GreedyBytes)
+    # 'MemData32' / LinearAdapter(32, Int8ub)
+    )
```
daniestevez, line 135: Can you point to some documentation describing this structure? What does the field `MemData32` contain?

python/telemetry/snet.py
```diff
@@ -168,10 +100,95 @@ def _decode(self, obj, context, path=None):
...
+    'CRC13' / BitsInteger(13),
+    'CRC5' / BitsInteger(5),
+    Padding(2)
+    )
+
+# For reference:
+#
+# To send mailbox messages
+# FCIDMajor = 56  (COM0)
+# FCIDSub = 34 || 134  (COM1)
```
daniestevez, line 128: This does not seem to match the `Switch()` below?

grc/satellites_snet_deframer.block.yml
```diff
@@ -9,9 +9,13 @@ parameters:
     options: ['True', 'False']
     option_labels: ['Yes', 'No']
 -   id: buggy_crc
-    label: Verbose
+    label: Buggy CRC
```
daniestevez, line 12: Good catch! Thanks.

grc/satellites_snet_deframer.block.yml
```diff
@@ -9,9 +9,13 @@ parameters:
...
     dtype: bool
     default: True
+-   id: userdata_only
+    label: UserData Only
+    dtype: bool
+    options: ['True', 'False']
```
daniestevez, line 18: What does this do? I don't see it used elsewhere.

daniestevez (commented):
It seems that the main change is the addition of a `MailboxTelemetry` telemetry packet, as an alternative option to `ADCSTelemetry` and `EPSTelemetry`. This could be achieved without rewriting this much code, and at the moment I don't see a reason that justifies this rewrite. Can you clarify?

grc/satellites_snet_deframer.block.yml
Gioyik: I was testing another thing, but yes, it should not be here.

python/telemetry/snet.py
Gioyik: my idea with this comment was to document, in case someone was curious, what were the `FCIDMajor` and `FCIDSub` values if someone happens to implement an `framer` (?). The values in the `Switch()` are the relevant IDs for the deframe action. Do you think this needs to be removed or put somewhere else instead?

python/telemetry/snet.py
Gioyik: There is no documentation available that I can reference, unfortunately. I got access to pieces where this was implemented, I can add more context about the values/types for each field.

python/telemetry/snet.py
Gioyik: you are right with this one, it was really not necessary to be moved, however, `PDUFrameHeader` originally `SNETFrameHeader` had to be moved because the introduction of the `Switch()` for `UserData` and its references to `ADCSTelemetry` and `EPSTelemetry`

python/telemetry/snet.py
Gioyik: I matched the names with what I saw (being named) in the implementation, so the deframer follows the name convention of how is defined internally and on university docs.

It's a nitpick from my side, probably yes, but I think when you get access to the SNET information and later check what is implemented in gr-satellites, you get an easier 1:1 relation to the different pieces and can connect the dots quickly.

I can roll back this change if it doesn't make sense, no problem for me.

python/telemetry/snet.py
Gioyik: That is a good point, as I don't have access or know what the structure is for the rest of the telemetry frames, most likely will not be the best way to parse in this way. As I am not entirely familiar with the construct library and this scenarios, what do you think would be the safe/best type to use to catch the `UserData` sections when they are not a case for the `Switch()`?

The only thing I know as a general rule for `UserData` is its size: 256 Bytes.

python/telemetry/snet.py
Gioyik: I agree. Is this comment different from what I put in the file?

I would like to extend my comment pointing the details you mention, but I want to make sure if I am reading your comment correctly.

python/snet_deframer.py
Gioyik: oh, that is a good point and I totally see how my changes are bias towards what, I feel is important, instead of a more generic approach that works for the overall. Thanks for the notes about the output to be a GNU Radio PDU, I definitely downgraded the importance of it.

Aside from this, do you think in the future, having parameters like this https://github.com/daniestevez/gr-satellites/blob/09b0d84ada7a168075cea751cdc08d589e6198f3/grc/satellites_snet_deframer.block.yml#L16 the return specific sections of the pdu make sense for the concept of deframer, or will it be always better to output everything and let an extra block or application parse the bits of interest?

daniestevez: I don't think that makes sense for a deframer. From the documentation:
> Deframers. Deframers implement the lower layer protocols related to frame boundary detection, descrambling, deinterleaving, FEC, error checking with a CRC code, etc. The output of a deframer are PDUs with the frames. Some examples are an AX.25 deframer and a CCSDS concatenated code deframer.

python/telemetry/snet.py
daniestevez: If the documentation is not publicly available, then I think the name change is not justified.

python/telemetry/snet.py
daniestevez: I think your comment matches your code, but my point is that:
1. The telemetry parser should probably support all the different types of frames transmitted by the satellite, so there wouldn't be unknown frames.
2. Documentation for all the frame types should generally be publicly available as per item 6.1 in [IARU Satellite Frequency Coordination request](https://www.iaru.org/wp-content/uploads/2020/06/iaru_amateur_satellite_coordination_request_v40.doc).
3. In the event that a payload doesn't match the formats supported by the telemetry parser, the parser should just ignore the payload (as it does now), since its goal is to parse fields into human readable format, and it cannot do so for frames of unsupported format.

python/telemetry/snet.py
daniestevez: See https://github.com/daniestevez/gr-satellites/pull/508#discussion_r1274123176

python/telemetry/snet.py
daniestevez: On a second look, is this actually correct? `BitsInteger(4)` is supposed to be a 4-bit integer, if I recall correctly. Then `BitsInteger(1)` is a 1-bit integer. Then we have `GreedyBytes`, but we're not byte aligned.

I guess that as it is this `Struct` is better than nothing, but it isn't very useful either. It just says that there is a timestamp, but doesn't indicate how to convert that to UTC time or whether it's seconds since boot or something, a memory data type, but without indicating what each type means, and finally a bunch of bytes of unknown content. That's why I asked for additional documentation.

python/telemetry/snet.py
daniestevez: Oh, so you mean that these are the FCIDMajor and FCIDMinor used in the uplink? It isn't clear from the comment (nor for your response, either; a framer could make frames identical to those transmitted by the satellite, or make uplink frames).

Gioyik: yes, these `FCID*` are for uplink and yes, makes sense the uplink can be added to the framer too, I'm using gr-satellites to deframe payloads from the satellite to ground only, so for that reason, I was only scoping the changes to downlink.

I will add it.

python/telemetry/snet.py
Gioyik: this was a bad take from my part, I switched it to catch the size of the `Payload` defined as 256 bytes

python/telemetry/snet.py
Gioyik: You are right on the first statement, it's an error from my side and I made a proper definition of the structure as a comment and later in the Struct. In addition to that, it shouldn't catch it as hex, the format of how you want to interpret the bits should be something outside the deframer.

As for the second point, I added more documentation from the logic I can extract, from the code I was provided. In addition to that, I fixed the links to the relevant docs of SNET at TUBerlin, however, none of them mention mailbox, and I am not aware of any documentation that explains its functionality nor implementation. I am doing my best to provide as much context with the little information I got around the topic, but I am sure the comments in the code today are accurate.

python/telemetry/snet.py
Gioyik: oh, I didn't know that, and now it makes sense why it was set to `default=Pass` in the first place.

I agree, and thanks for the information/context.

python/telemetry/snet.py
Gioyik: I liked the new docs at the top of the file where the link was broken, could you please check and see if the rename makes sense now?

still, I don't have a problem rolling back these specific changes.

python/snet_deframer.py
Gioyik: thanks for the clarification, I rolled back this change.

python/telemetry/snet.py
daniestevez: Okay for the `SNETFrameHeaderExtension` -> `PDUFrameHeaderExtension` and `SNETFrameExtension` -> `PDUFrameHeader` rename, but that's the only thing that should show up in the diff.

python/telemetry/snet.py
```diff
@@ -168,10 +102,105 @@ def _decode(self, obj, context, path=None):
...
+    'UserData' / Switch(lambda c: (c.FCIDMajor, c.FCIDSub), {
         (0, 0): ADCSTelemetry,
         (9, 0): EPSTelemetry,
-    }, default=Pass))
+        (54, 17): MailboxReceiveTelemetry,  # COM0
+        (54, 117): MailboxReceiveTelemetry,  # COM1
+        (56, 34): MailboxSendTelemetry,  # COM0
+        (56, 134): MailboxSendTelemetry,  # COM1
+    },
+      default=Pass))
```
daniestevez, line 202: The user data should not be part of the header, right?

python/snet_deframer.py
```diff
@@ -11,9 +11,10 @@
 from gnuradio import gr
 import numpy as np
 import pmt
+from construct import *
```
daniestevez, line 14: I think this import is not needed.

python/snet_deframer.py
```diff
@@ -11,9 +11,10 @@
...
 
 from .bch15 import decode_bch15
-from .telemetry.snet import LTUFrameHeader
+from .telemetry.snet import LTUFrameHeader, PDUFrameHeader
```
daniestevez, line 17: If we don't use the `PDUFrameHeader` below, then we do not need to change the import.

daniestevez (changes requested):

python/telemetry/snet.py
Gioyik: moved it back

python/telemetry/snet.py
Gioyik: yes, it is separated now

Gioyik:
@daniestevez I think I addressed most of your comments and requests for changes, could you please give it a check again?

This time, I added an Adapter for the `MemData32` and changed the `norad` values accordingly to this reference: https://web.archive.org/web/20200106160215/http://dk3wn.info:80/blog/satelliten/s-net/

python/satyaml/S-NET_A.yml
```diff
@@ -1,5 +1,5 @@
 name: S-NET A
-norad: 43188
+norad: 43186
```
daniestevez, line 2: Are you sure about this? The DK3WN reference doesn't say whether S-NET A is 43186. Celestrak lists 43188 for S-NET A.

python/telemetry/snet.py
```diff
@@ -168,10 +167,57 @@ def _decode(self, obj, context, path=None):
...
+    'MemDataType' / BitsInteger(8),
+    'MemData32' / MemData32
+    )
+
+MailboxSendTelemetry = Struct(
+    'MemDataType' / BitsInteger(8),
+    'MemData32' / BitsInteger(256)
+    )
+
+Payload = BitStruct(
```
daniestevez, line 210: Why this should be a `BitStruct` rather than a `Struct`? Also, we probably don't need a single-field `Struct`. This being a field of the `snet` `Struct` as it was before would be better.

python/telemetry/snet.py
```diff
@@ -168,10 +167,57 @@ def _decode(self, obj, context, path=None):
...
+            (obj.upper() + ' '*8)[:8], encoding='ascii')])
+
+    def _decode(self, obj, context, path=None):
+        # do we need to do right shift here?
+        return str(bytes([x for x in bytes(
+            (obj.upper() + ' '*8)[:8], encoding='ascii')]),
+                   encoding='ascii').strip()
+
+
+MemData32 = MemData32Adapter(BitsInteger(256))
```
daniestevez, line 197: Why can't we use something like `Bytes(32)` here?

python/telemetry/snet.py
```diff
@@ -168,10 +167,57 @@ def _decode(self, obj, context, path=None):
...
+
+MailboxReceiveTelemetry = Struct(
+    'SysTime' / BitsInteger(32),
```
daniestevez, line 200: Why `BitsInteger(32)` instead of something like `Int32sb`?

python/telemetry/snet.py
Gioyik: no hard reason why using `BitStruct` instead of `Struct`. I am fine taking the change of the Payload inside the `snet` `Struct` instead.

daniestevez: I don't think it is necessary to use `BitStruct` here. All the fields are `Struct`'s, so we don't need the [`Bitwise` transform](https://construct.readthedocs.io/en/latest/api/core.html#construct.core.Bitwise).

python/satyaml/S-NET_A.yml
Gioyik: I reverted the changes on these files, they were right before

python/telemetry/snet.py
Gioyik: no strong reason why, I have being using `Bits*` mostly, but `Int32sb` works too

daniestevez:
I'm going to close this PR, since the fix for the SNET Deframer GRC block has already been merged and the other changes in the PR were pending testing, code changes, etc. The PR has been stale for almost a year now. Feel free to reopen with updates.

PR #509, opened 2023-07-25
Update UmKA-1.yml

janvgils:
The satellite can run with USP and AX.25 G3RUH framing and therefor also created the USP transmitters


daniestevez:
Does the USP also transmit AX.25 frames, or does it need a different telemetry parser (maybe `unknown`)?

janvgils:
It also uses AX.25

daniestevez:
Great. Merging.

PR #510, opened 2023-08-01
Correct options label and documentation for pdu head/tail

jganser:
**Correct options label and documentation**

Tail- was actually corresponding to tail+ in Unix
Head+ as it was Head- was wrongly named in options label, though correctly named in the documentation.

Looked it up in the cc file, where naming was correct and matching the UNIX behavior.

As can also be seen by my bash tryouts.
```bash
echo "123456789" | tail -c 3
89

echo "123456789" | tail -c -3
89

echo "123456789" | tail -c +3
3456789
```

```bash
echo "123456789" | head -c 3
123⏎

echo "123456789" | head -c -3
1234567⏎ 

echo "123456789" | head -c +3
123⏎ 
```

### DIFF
--- a/grc/satellites_pdu_head_tail.block.yml
+++ b/grc/satellites_pdu_head_tail.block.yml
@@ -8,7 +8,7 @@ parameters:
     dtype: enum
     default: '0'
     options: ['0', '1', '2', '3']
-    option_labels: ['Head', 'Head-', 'Tail', 'Tail-']
+    option_labels: ['Head', 'Head-', 'Tail', 'Tail+']
 -   id: num
     label: Number
     dtype: int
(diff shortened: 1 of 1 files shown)


jganser:
Found the mismatch in the gr310 build on Ubuntu gr-satellites version 5.3.0-2

daniestevez:
Thanks! Indeed reviewing the [C++ code](https://github.com/daniestevez/gr-satellites/blob/267276336b77db8bc925657d17720c8e8f9a0727/lib/pdu_head_tail_impl.cc#L68) I find that these were correctly named in there, so I don't know why I goofed up when writing the GRC file.

PR #520, opened 2023-11-29
Add IMPULSE-1 satyaml

kng:
I have been looking at the data and it seems like the first byte in the frames are actually frame length, this could perhaps be checked and also removed from the resulting output.


kng:
I'll retract this until the discussions settle down.

PR #526, opened 2023-12-31
Fix GRC template to use correct type for fixedlen_to_pdu instantiation

### LINKED ISSUE #525: Wrong vector type instantiation in Fixed Length To PDU block
Hi,

I tried to use the Fixed Length To PDU block with a custom block that produces a stream of floats as output (custom block -> fixedlen_to_pdu block). GRC is happy with the float to float connection, but when starting the flowgraph I get a  `ValueError: itemsize mismatch: custom_block0:0 using 4, fixedlen_to_pdu0:0 using 8`.

The `fixedlen_to_pdu block`  is instantiated with `satellites.fixedlen_to_pdu(gr.sizeof_float, 'frame_score', (2048+32), False)` (as generated by GRC). `gr.sizeof_float` is an `int`, but the C++ block itself expects a `gr::type::vector_type` which is an enum (and thus it can be implicitly constructed from int without any error) but the underlying enum values do not match the size of the type.

```
>>> from gnuradio import gr
>>> vt = gr.sizeof_float
>>> vt
4
>>> type(vt)
<class 'int'>
>>> gr.pdu.itemsize(vt)
8
>>> gr.pdu.type_from_pmt(gr.pdu.make_pdu_vector(vt, 0, 2))
<vector_type.complex_t: 4>
```

```
>>> from gnuradio import gr
>>> vt = gr.types.vector_type(gr.types.float_t)
>>> vt
<vector_type.float_t: 3>
>>> type(vt)
<class 'gnuradio.gr.gr_python.types.vector_type'>
>>> gr.pdu.itemsize(vt)
4
>>> gr.pdu.type_from_pmt(gr.pdu.make_pdu_vector(vt, 0, 2))
<vector_type.float_t: 3>
```

As you can see `gr.sizeof_float` correctly reports that a float is 4 bytes in GnuRadio, but enum value number 4 in `gr::type::vector_type` corresponds to `complex_t`, which in turn has a size of 8 bytes, not 4.

```
>>> gr.types.vector_type(3)
<vector_type.float_t: 3>
>>> gr.pdu.itemsize(int(3))
4
>>> gr.types.vector_type(4)
<vector_type.complex_t: 4>
>>> gr.pdu.itemsize(int(4))
8
```

I believe the correct way to instantiate fixedlen_to_pdu is thus with `gr.types.vector_type(gr.types.float_t)` instead of `gr.sizeof_float`. I will create a pull request to fix this.

v1tzl1:
fixes #525

### DIFF
--- a/grc/satellites_fixedlen_to_pdu.block.yml
+++ b/grc/satellites_fixedlen_to_pdu.block.yml
@@ -6,10 +6,9 @@ parameters:
 -   id: type
     label: Type
     dtype: enum
-    options: [complex, float, int, short, byte]
+    options: [complex, float, byte]
     option_attributes:
-        size: [gr.sizeof_gr_complex, gr.sizeof_float, gr.sizeof_int, gr.sizeof_short,
-               gr.sizeof_char]
+        vector_type: [satellites.grtypes.complex_t, satellites.grtypes.float_t, satellites.grtypes.byte_t]
     hide: part
 -   id: syncword_tag
     label: Syncword tag
@@ -35,7 +34,8 @@ outputs:
 templates:
     imports: |-
         import satellites
-    make: satellites.fixedlen_to_pdu(${type.size}, ${syncword_tag}, ${packet_len}, ${pack if type == 'byte' else False})
+        import satellites.grtypes
+    make: satellites.fixedlen_to_pdu(${type.vector_type}, ${syncword_tag}, ${packet_len}, ${pack if type == 'byte' else False})
 
 documentation: |-
     Extracts PDUs of fixed length according to syncword tags


grc/satellites_fixedlen_to_pdu.block.yml
```diff
@@ -35,7 +34,8 @@ outputs:
 templates:
     imports: |-
         import satellites
-    make: satellites.fixedlen_to_pdu(${type.size}, ${syncword_tag}, ${packet_len}, ${pack if type == 'byte' else False})
+        import satellites.grtypes
+    make: satellites.fixedlen_to_pdu(gr.types.vector_type(${type.vector_type}), ${syncword_tag}, ${packet_len}, ${pack if type == 'byte' else False})
```
daniestevez, line 38: Is the explicit conversion to `gr.types.vector_type` necessary here? Comparing with Tagged Stream to PDU, I see that it [uses the types from `gr.types` directly](https://github.com/gnuradio/gnuradio/blob/main/gr-pdu/grc/pdu_tagged_stream_to_pdu.block.yml#L11) (and its constructor [also has a `gr::types::vector_type` parameter](https://github.com/gnuradio/gnuradio/blob/main/gr-pdu/include/gnuradio/pdu/pdu_to_tagged_stream.h)).

I suspect that this would also work without the explicit conversion to `gr.types.vector_type` and reduces the chances of breaking things under GR 3.9 or 3.8.

v1tzl1: I changed the PR accordingly. I can confirm it still works for GnuRadio 3.10, so the GnuRadio pybinds seem to have some explicit conversion from the types enum to vector_type.

I don't have a working GnuRadio 3.8 or 3.9 ready to test it over there, sorry. The whole point of me using gr-satellites over my old outdated custom OOT was to start on a blank and modern canvas.

daniestevez:
This looks good now. Merging.

PR #544, opened 2024-04-14
Fix varlen blocks wrong args type for gnuradio-3.10.

b-o-r-m-a-l-e-y:
Fixes error `Expression <endianness_t.GR_MSB_FIRST: 0> is invalid for type 'int'.` in varlen_packed_tagger, tested with gr-3.10.

### DIFF
--- a/grc/satellites_varlen_packet_framer.block.yml
+++ b/grc/satellites_varlen_packet_framer.block.yml
@@ -15,9 +15,9 @@ parameters:
     default: '12'
 -   id: endianness
     label: Endianness
-    dtype: int
-    default: int(gr.GR_MSB_FIRST)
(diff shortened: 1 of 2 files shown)


daniestevez:
Thank you! Do you know if the fix is also applicable to GNU Radio 3.9? These two blocks were originally written for GNU Radio 3.8, and this problem looks like a SWIG vs pybind11difference in the type of the expected arguments.

b-o-r-m-a-l-e-y:
> Thank you! Do you know if the fix is also applicable to GNU Radio 3.9? These two blocks were originally written for GNU Radio 3.8, and this problem looks like a SWIG vs pybind11difference in the type of the expected arguments.

You are welcome :) As far as I can remember this fix also forks with 3.9.

daniestevez:
Great. I'll merge this and also apply the fix to the `main-3.9` branch.

PR #550, opened 2024-06-17
Update BISONSAT.yml

janvgils:
Now that the TLE are publicly shared by SpaceTrack the norad is swapped


daniestevez:
Thanks! Merging.

PR #554, opened 2024-06-24
Add deviation to MCUBED-2

kng:
Observation [9702028](https://network.satnogs.org/observations/9702028/)
dd bs=$((4*57600)) if=iq_9702028_57600.raw of=mcubed2_cut.raw skip=357 count=1
gr_satellites mcubed-2 --iq --rawint16 mcubed2_cut.raw --samp_rate 57600 --dump_path dump --deviation 2400 --disable_dc_block
mcubed2_dev2400

### DIFF
--- a/python/satyaml/MCUBED-2.yml
+++ b/python/satyaml/MCUBED-2.yml
@@ -8,6 +8,7 @@ transmitters:
     frequency: 437.480e+6
     modulation: FSK
     baudrate: 9600
+    deviation: 2400
     framing: AX.25 G3RUH
     data:
     - *tlm


daniestevez:
This looks right, since we already discussed this particular satellite. However, for future plots can you use a continuous line rather than dots (`-` instead of `.`) and zoom in so that individual symbols are visible? Otherwise it's difficult to see even if this is a proper FSK waveform.

daniestevez (approved):

kng:
mcubed2_dev2400_lines

PR #556, opened 2024-06-24
Add deviation to TIGRISAT

kng:
TIGRISAT (40043)
Observation [9691434](https://network.satnogs.org/observations/9691434/)
dd bs=$((4*57600)) if=iq_9691434_57600.raw of=tigrisat_cut.raw skip=510 count=1
gr_satellites tigrisat --iq --rawint16 tigrisat_cut.raw --samp_rate 57600 --dump_path dump --deviation 2400 --disable_dc_block
tigrisat_dev2400

### DIFF
--- a/python/satyaml/TIGRISAT.yml
+++ b/python/satyaml/TIGRISAT.yml
@@ -8,6 +8,7 @@ transmitters:
     frequency: 435.000e+6
     modulation: FSK
     baudrate: 9600
+    deviation: 2400
     framing: AX.25 G3RUH
     data:
     - *tlm


daniestevez:
Here the deviation looks slightly larger than 2400 Hz (the same also happens in #554). I guess that is makes sense to round it to 2400 Hz, since that would be (G)MSK, but I don't think we can rule out that the transmitter is using a slightly larger deviation. In any case, 2400 Hz will work better than the default 5000 Hz.

daniestevez (approved):

kng:
tigrisat_dev2400_lines

PR #562, opened 2024-06-27
Update deviation on EIRSAT-1

kng:
EIRSAT-1 (58472)
Observation [9494233](https://network.satnogs.org/observations/9494233/)
dd bs=$((4*57600)) if=iq_9494233_57600.raw of=eirsat1_cut.raw skip=366 count=5
gr_satellites EIRSAT-1.yml --iq --rawint16 eirsat1_cut.raw --samp_rate 57.6e3 --dump_path dump --disable_dc_block --deviation -3500
run with old deviation removed from satyaml, result 1.15 to -0.85 
eirsat1_dev3500

Here's with the old -2400 deviation for reference:
eirsat1_dev2400

### DIFF
--- a/python/satyaml/EIRSAT-1.yml
+++ b/python/satyaml/EIRSAT-1.yml
@@ -13,7 +13,7 @@ transmitters:
     RS basis: dual 
     RS interleaving: 4
     convolutional: CCSDS uninverted
-    deviation: -2400
+    deviation: -3500
     data:
     - *tlm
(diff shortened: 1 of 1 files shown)


daniestevez:
Even though the team mentioned that the satellite uses 9k6 GMSK in #307, the deviation seems closer to 3500 Hz than 2400 Hz. This looks good. Merging.

PR #565, opened 2024-07-03
Fix USP AX25 crop block issue in Gnuradio.

b-o-r-m-a-l-e-y:
Hello, GNU Radio can`t use blocks with dots in name, I get error when using it in gnuradio-companion. PR fixes this.


daniestevez:
Thank you! This makes sense. I don't know how the mistake went unnoticed.

PR #572, opened 2024-07-06
Update norad in AHMAT-1

kng:
https://db.satnogs.org/satellite/ULWR-5929-7873-9279-5662

N2YO says it is https://www.n2yo.com/satellite/?s=57208  but Celestrak and SatNOGS have identified that one as SITRO-AIS-11


daniestevez (approved):
Looks good.

PR #574, opened 2024-07-06
Update norad in IRIS

kng:
Decayed.
https://www.n2yo.com/satellite/?s=57315#results
https://db.satnogs.org/satellite/YUVU-2466-2581-7667-8758


daniestevez (approved):
Not completely clear that the ID is correct, because Celestrak lists it as OBJECT VT, and SatNOGS as "followed NORAD ID". In any case it doesn't matter much, since it's already decayed. Approved.

PR #575, opened 2024-07-06
Update norad in KS-1Q

kng:
Old IS is decayed: http://celestrak.org/satcat/table-satcat.php?CATNR=41845&PAYLOAD=1&MAX=500

http://celestrak.org/satcat/table-satcat.php?CATNR=41847&PAYLOAD=1&MAX=500
https://db.satnogs.org/satellite/WMGA-3000-1529-8035-2634
another name but beacon frequency seems to match https://www.n2yo.com/satellite/?s=41847#results

### DIFF
--- a/python/satyaml/KS-1Q.yml
+++ b/python/satyaml/KS-1Q.yml
@@ -1,5 +1,5 @@
 name: KS-1Q
-norad: 41845
+norad: 41847
 data:
   &tlm Telemetry:
     telemetry: csp


daniestevez:
KS-1Q is permanently attached to the CZ-11 rocket upper stage, together with another satellite called CAS-2T. [More info](https://destevez.net/2016/11/about-ks-1q/). Celestrak is listing 41847 as this object, and 41845 as a decayed free-flying satellite. I don't know why I had 41845 for this one. Maybe the TLEs got swapped at some point or the TLE lottery was wrong. This looks good. Approved.

daniestevez (approved):

PR #576, opened 2024-07-06
Update norad in QUARMAN

kng:
Decayed
http://celestrak.org/satcat/table-satcat.php?CATNR=45263&PAYLOAD=1&MAX=500
https://db.satnogs.org/satellite/XJKP-5229-8707-4977-2749
https://www.n2yo.com/satellite/?s=45263#results


daniestevez:
If Celestrak says so, it looks good. Approved.

daniestevez (approved):

PR #592, opened 2024-07-13
Add deviation to TTU-100

kng:
TTU-100 (46312)
Observation [9841235](https://network.satnogs.org/observations/9841235/)
dd bs=$((4*48000)) if=iq_9841235_48000.raw of=cut.raw skip=254 count=1
gr_satellites ttu-100 --iq --rawint16 cut.raw --samp_rate 48000 --dump_path dump --disable_dc_block --deviation 3000
ttu100_dev3000
plot with histogram, unfortunately I haven't figured out how to share x-axis between the two to only show histogram with the current zoom of the plot, so the hist is for one second and includes all the noise before/after the frame.

### DIFF
--- a/python/satyaml/TTU-100.yml
+++ b/python/satyaml/TTU-100.yml
@@ -10,6 +10,7 @@ transmitters:
     frequency: 435.450e+6
     modulation: FSK
     baudrate: 9600
+    deviation: 3000
     framing: AX.25 G3RUH
     data:
     - *tlm


daniestevez:
I don't think it's possible (or at least reasonably easy) to get the histogram to recompute when you zoom in on the time domain plot. It's not really the same as syncing the axes of two plots, which can be done (here you could sync the amplitude axis on both plots, for instance).

kng:
ah, I figured :P well, the current script is [here](https://gist.github.com/kng/232d3e8badb77f1e0946814896ce4323), would this fit in gr-satellites apps ? might need some argparse and polishing perhaps.

daniestevez:
Everything that goes in `apps` is installed to the system PATH, so it needs a man page (which makes me realize that `erminaz_ssdv` is missing a man page). Also, the name of the script should be well descriptive (since the script ends up being a system-wide command). This is required by Debian and other distros.

If this script is only intended for people who are expected to have a checkout of the gr-satellites source, then it can go into `tools`. The scripts there don't get installed, so the requirements are much laxer.

PR #603, opened 2024-07-14
Add deviation to GREENCUBE

kng:
GREENCUBE (53106)
Observation [9847373](https://network.satnogs.org/observations/9847373/)
dd bs=$((4*48000)) if=iq_9847373_48000.raw of=cut.raw skip=252 count=2
gr_satellites GREENCUBE_12.yml --iq --rawint16 cut.raw --samp_rate 48000 --dump_path dump --disable_dc_block --deviation 500
greencube_b1200_dev500

Observation [8842104](https://network.satnogs.org/observations/8842104/)
dd bs=$((4*48000)) if=iq_8842104_48000.wav of=cut.raw skip=693 count=10
gr_satellites GREENCUBE_3.yml --iq --rawint16 cut.raw --samp_rate 48000 --dump_path dump --disable_dc_block --deviation 125
greencube_b300_dev125

The 300 baud one was quite hard, I only had a older observation of the 300 baud transmission and it was off by something like 280Hz so when reducing the deviation to it's proper value the signal was out of the passband so the waveform was useless.
I had to run it through a flowgraph to correct the frequency, that is why it is using a .wav

### DIFF
--- a/python/satyaml/GREENCUBE.yml
+++ b/python/satyaml/GREENCUBE.yml
@@ -8,6 +8,7 @@ transmitters:
     frequency: 435.310e+6
     modulation: FSK
     baudrate: 300
+    deviation: 125
     framing: AX100 ASM+Golay
     data:
     - *tlm
@@ -22,6 +23,7 @@ transmitters:
     frequency: 435.310e+6
     modulation: FSK
     baudrate: 1200
+    deviation: 500
     framing: AX100 ASM+Golay
     data:
     - *tlm


daniestevez:
Thanks! The [Greencube team says that it uses GMSK](https://www.s5lab.space/index.php/decoding-ledsat-2/), but according to your measurements this is clearly not the case. GMSK would have `deviation = baudrate / 4`.

K4KDR:
For a long time now, the vast majority of the downlink traffic from GreenCube was either 1200 baud or (less frequently) 300 baud packets digipeated from stations making 2-way contacts.   A special version of SoundModem for GreenCube was created by UZ7HO and, while USB is the ONLY option for uplink packets, the downlink packets can actually be decoded by either the USB -or- FM version of this custom UZ7HO SoundModem.   (http://uz7.ho.ua/modem_beta/greentnc.zip)

Screenshot from 2024-07-15 14-25-15

PR #655, opened 2024-08-16
Add CUAVA-2 and WS-1 satellites

patrickoppel:
As a representative of the University of Sydney team, I'd like to add the CUAVA-2
and Waratah Seed-1 CubeSats. 

Please let me know if this is not enough for an endorsement of
commercial satellites.

### DIFF
--- a/python/satyaml/WS-1.yml
+++ b/python/satyaml/WS-1.yml
@@ -1,0 +1,15 @@
+name: WS-1
+alternative_names: 
+  - Waratah Seed-1
+norad: 98849
+data:
+  &tlm Telemetry:
+    telemetry: ax25
+transmitters:
+  1k2 BPSK downlink:
+    frequency: 400.65e+6
+    modulation: BPSK
+    baudrate: 1200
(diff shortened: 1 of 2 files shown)


daniestevez (changes requested):
Thanks. This looks good. Enough endorsement, since the pull request comes from a team representative.

Can you add a newline at the end of each SatYAML file, add an alternative name for WS-1 and squash into a single commit?

The syntax for the alternative name would be something like
```
alternative_names:
  - Waratah Seed-1
```
usually placed under `name:`.

patrickoppel:
Added new line, alternative_name and squashed

python/satyaml/WS-1.yml
```diff
@@ -0,0 +1,14 @@
+name: WS-1
+alternative_names: Waratah Seed-1
```
daniestevez, line 2:
```suggestion
alternative_names:
  - Waratah Seed-1
```

daniestevez:
Looks good now. Merging.

PR #658, opened 2024-08-17
Add Hades deframer to SatYAML and core flowgraph

kng:


daniestevez (changes requested):
Thanks. The changes look good.

This also needs to add `Hades` to the list of framings in the [SatYAML documentation](https://github.com/daniestevez/gr-satellites/blob/main/docs/source/satyaml.rst). Something like

```
- `Hades`, the custom framing used by HADES-D.
```

after `QUBIK` would be good.

kng (commented):
done.

daniestevez (approved):
Looks good.

PR #662, opened 2024-09-07
telemetry: Change line endings to LF

Akira25:
While packaging gr-satellites into an rpm, the linter complained on these files. As all other files use LF as line-ending, changing these to be consistent with the rest of the project.


daniestevez:
Good find! I didn't notice that these files were CRLF when they were submitted, since the Github web UI doesn't seem to show any indication and we don't check in CI. Everything in gr-satellites should indeed be LF. Merging.

PR #663, opened 2024-09-10
Change savefig dir in waveform_plot

kng:
This sets the working dir to the same as where the script was launched, and hopefully closer to where the resulting images is to be saved.
https://matplotlib.org/stable/users/explain/customizing.html#the-default-matplotlibrc-file
Closing the plot might alleviate the issue where displaying the plot twice in some older versions.


daniestevez:
Looks good. Merging.

PR #664, opened 2024-09-21
Update Bluebird_01.yml

janvgils:
Updated Object ID

### DIFF
--- a/python/satyaml/Bluebird_01.yml
+++ b/python/satyaml/Bluebird_01.yml
@@ -1,5 +1,5 @@
 name: Bluebird 01
-norad: 98804
+norad: 61047
 data:
   &tlm Telemetry:
     unknown


janvgils:
Not in line with celestrak so close pull request

daniestevez:
Thanks for keeping an eye on this. Reading [this](https://community.libre.space/t/asts-bluebird-1-5/12255/17) it seems that there might be some confusion with these satellites, because the FCC grant gave 5 UHF frequencies in increasing order and we had assumed that they were assigned to the Bluebird-1 to Bluebird-5 satellites also in increasing order, but this might not be the case. 

The correct approach in this case is to wait until Space-Track/Celestrak have NORAD-IDs assigned for all these satellites (these objects will be claimed by AST, who know what they're doing and can tell their satellites apart). Then we can play "TLE lottery" in the community to match each NORAD ID with its transmit frequency.

For gr-satellites SatYAML the frequencies are not too important, because they're present only for documentation purposes. Matching the satellite name and NORAD ID is more important, because the NORAD ID is used for telemetry submission. So as a first step we should strive to make the NORAD IDs in the SatYAMLs to match those on Celestrak as they keep appearing.

PR #678, opened 2024-10-28
Added new deframer for GEOSCAN satellite platform

Foxiks:
Added new deframer for GEOSCAN satellite platform. New packet length is 74 bytes. I changed the block by adding a parameter to select packet length for old and new satellite platform

### DIFF
--- a/grc/components/deframers/satellites_geoscan_deframer.block.yml
+++ b/grc/components/deframers/satellites_geoscan_deframer.block.yml
@@ -41,6 +41,7 @@ documentation: |-
         PDUs with the deframed GEOSCAN packets
 
     Parameters:
+        Frame size: size of the decode frame in bytes
         Syncword threshold: number of bit errors to allow in syncword detection
         Command line options: options to pass down to the block, following the syntax of the gr_satellites command line tool
 
--- a/grc/components/deframers/satellites_geoscan_deframer.block.yml
+++ b/grc/components/deframers/satellites_geoscan_deframer.block.yml
@@ -12,7 +12,7 @@ parameters:
     dtype: string
     default: '""'
     hide: part
--   id: packet_length
+-   id: frame_size
     label: Frame length (bytes)
     dtype: int
     default: 66
@@ -27,7 +27,7 @@ outputs:
 
 templates:
     imports: import satellites.components.deframers
-    make: satellites.components.deframers.geoscan_deframer(packet_length = ${packet_length}, syncword_threshold = ${threshold}, options=${options})
+    make: satellites.components.deframers.geoscan_deframer(frame_size = ${frame_size}, syncword_threshold = ${threshold}, options=${options})
 
 documentation: |-
     Deframes GEOSCAN custom packets
--- a/grc/components/deframers/satellites_geoscan_deframer.block.yml
+++ b/grc/components/deframers/satellites_geoscan_deframer.block.yml
@@ -12,10 +12,10 @@ parameters:
     dtype: string
     default: '""'
     hide: part
--   id: new_geoscan_platform
-    label: New GEOSCAN Platform deframer
+-   id: packet_length
+    label: Frame length (bytes)
     dtype: int
-    default: 0
+    default: 66
 
 inputs:
 -   domain: stream
@@ -27,10 +27,10 @@ outputs:
 
 templates:
     imports: import satellites.components.deframers
-    make: satellites.components.deframers.geoscan_deframer(new_geoscan_platform = ${new_geoscan_platform}, syncword_threshold = ${threshold}, options=${options})
+    make: satellites.components.deframers.geoscan_deframer(packet_length = ${packet_length}, syncword_threshold = ${threshold}, options=${options})
 
 documentation: |-
-    Deframes AALTO-1 custom packets
+    Deframes GEOSCAN custom packets
 
     The frames are transmitted by a Texas Intruments CC1125 transceiver with a PN9 scrambler and the built-in CRC.
 
--- a/grc/components/deframers/satellites_geoscan_deframer.block.yml
+++ b/grc/components/deframers/satellites_geoscan_deframer.block.yml
@@ -12,6 +12,10 @@ parameters:
     dtype: string
     default: '""'
     hide: part
+-   id: new_geoscan_platform
+    label: New GEOSCAN Platform deframer
+    dtype: int
+    default: 0
 
 inputs:
 -   domain: stream
@@ -23,7 +27,7 @@ outputs:
 
 templates:
     imports: import satellites.components.deframers
-    make: satellites.components.deframers.geoscan_deframer(syncword_threshold = ${threshold}, options=${options})
+    make: satellites.components.deframers.geoscan_deframer(new_geoscan_platform = ${new_geoscan_platform}, syncword_threshold = ${threshold}, options=${options})
 
 documentation: |-
     Deframes AALTO-1 custom packets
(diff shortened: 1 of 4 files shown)


python/components/deframers/geoscan_deframer.py
```diff
@@ -45,10 +45,15 @@ def __init__(self, syncword_threshold=None, options=None):
 
         if syncword_threshold is None:
             syncword_threshold = self.options.syncword_threshold
+        
+        if new_geoscan_platform:
+            self.packlen=74
+        else:
+            self.packlen=66
```
daniestevez, line 52: I'm wondering if instead of having a boolean `new_geoscan_platform` parameter we should have a `packlen` parameter. This seems better moving forward in case the packet length is changed yet again.

Also, there is no reason to create a class member `self.packlen` here. This could have been assign to a local variable `packlen` instead.

daniestevez (changes requested):
The new parameter is not reachable through SatYAML. A way to specify this framing should be added in [satyaml.py](https://github.com/daniestevez/gr-satellites/blob/main/python/satyaml/satyaml.py), documented in [satyaml.rst](https://github.com/daniestevez/gr-satellites/blob/main/docs/source/satyaml.rst), and hooked up in [gr_satellites_flowgraph.py](https://github.com/daniestevez/gr-satellites/blob/main/python/core/gr_satellites_flowgraph.py).

Do you have example recordings of the new packet format?

Is this something to which the current GEOSCAN satellites will be switching in flight or is it just for upcoming satellites? I'm trying to figure out what the required changes to existing SatYAML files are.

Foxiks:
Actually the new packet length will only be used on the new satellites, but StartoSat TK-1 will still have the old packet length, so I made the switch via a parameter, but maybe it's worth just dedicating a new block for this?

daniestevez:
Your idea to use the same block is preferable, because it avoids unnecessary code duplication.

There is also `GEOSCAN-EDELVEIS.yml` using the GEOSCAN deframer. Will this satellite maintain the old packet length too?

Foxiks:
GEOSCAN-EDELVEIS.yml is applicable for StratoSat TK-1, and new satellites can also use it, only it is necessary to specify the packet length not 66, but 74 bytes, and everything else remains the same

Foxiks:
The GEOSCAN-EDELVEIS satellite is long gone, but this yml file is suitable for StratoSat TK-1

daniestevez:
I wasn't aware that GEOSCAN-EDELVEIS had decayed. StratoSat TK-1 has its own SatYAML file [STRATOSAT-TK1.yml](https://github.com/daniestevez/gr-satellites/blob/main/python/satyaml/STRATOSAT-TK1.yml). Each satellite supported by gr-satellites has its own SatYAML file.

Foxiks:
Excellent! The thing is that I use separate deframers inside GR and did not pay attention to SatYML. Tell me what I should change or add to add support for the new GEOSCAN satellite platform?

daniestevez:
I outlined the required changes in

> The new parameter is not reachable through SatYAML. A way to specify this framing should be added in [satyaml.py](https://github.com/daniestevez/gr-satellites/blob/main/python/satyaml/satyaml.py), documented in [satyaml.rst](https://github.com/daniestevez/gr-satellites/blob/main/docs/source/satyaml.rst), and hooked up in [gr_satellites_flowgraph.py](https://github.com/daniestevez/gr-satellites/blob/main/python/core/gr_satellites_flowgraph.py).

Can you take a look at existing code in those files and let me know if you have questions?

Foxiks:
Yes, I will try to work with this.

Foxiks:
I added a parameter for variable packet length, because I was informed that on the new GEOSCAN satellite platform it is possible to change the packet length up to 255 bytes

python/components/deframers/geoscan_deframer.py
```diff
@@ -73,3 +72,7 @@ def add_options(cls, parser):
             '--syncword_threshold', type=int,
             default=cls._default_sync_threshold,
             help='Syncword bit errors [default=%(default)r]')
+        parser.add_argument(
+            '--packet_length', type=int,
+            default=cls._default_packet_length,
+            help='Frame length in bytes [default=%(default)r]')
```
daniestevez, line 78: Packet length shouldn't be handled via command line options. The SatYAML file should contain `framing`s for all the packet lengths that the satellite can transmit. The user should not need to specify a packet length with command line options.

Foxiks: The thing is that we don’t yet know exactly what the lengths might be, they can be completely different and change over time.

daniestevez: If the packets do not include a length field this means that the receiver must know the packet length ahead of time. Therefore, the SatYAML file will need to be updated any time that the satellite transmitter configuration changes. This is unfortunate, because of the need to keep updating this file, but the responsibility of using the correct packet length cannot lie within the users of gr-satellites. Command line parameters are to adjust receiver parameters that are not directly related to the transmitter modulation and coding configuration, and that might need to be adjusted for different SNRs, clock stabilities, etc. (examples: loop bandwidths and syncword thresholds).

There is already a `frame size` parameter in SatYAML ([example](https://github.com/daniestevez/gr-satellites/blob/main/python/satyaml/EIRSAT-1.yml#L12)). This is used by the CCSDS deframers and also by other deframers, such as `SMOG-P RA`. The same `frame size` parameter should also be used here with the GEOSCAN deframer.

Foxiks: To be honest, I don't quite understand how to add the frame size field to the yml file code. How can I do this?

Foxiks: I think you might want to take a look at this

daniestevez: At what? The changes in 17cddff look good, but there are still some comments pending (mostly about documentation) and pycodestyle doesn't pass.

grc/components/deframers/satellites_geoscan_deframer.block.yml
```diff
@@ -12,6 +12,10 @@ parameters:
     dtype: string
     default: '""'
     hide: part
+-   id: frame_size
```
daniestevez, line 15: New parameter missing in documentation below.

daniestevez:
Besides the open conversations, I'm missing a way to test these changes. For instance, a SatYAML file and sample recording for one of the new satellites.

python/satyaml/STRATOSAT-TK1.yml
```diff
@@ -12,5 +12,6 @@ transmitters:
      baudrate: 9600
      deviation: 4800
      framing: GEOSCAN
+     frame size: 66
```
daniestevez, line 15: Can you also add the same to `GEOSCAN-EDELVEIS.yml`?

Foxiks: GEOSCAN-EDELVEIS.yml I think it's worth deleting this file. It's no longer relevant.

daniestevez: We keep SatYAML files for decayed satellites. People can use them to decode recorded IQ data.

Foxiks: Okay, I'll add that. I'll also provide you with IQ to process the new GEOSCAN satellite platform.

Foxiks: You can test it on this IQ recording https://drive.google.com/file/d/1YBSFlFV70HCjQ42WSrACk6HK5Wbro2nC/view?usp=sharing This signal is made on a test model of the satellite. You can make a similar satyml file for this, just specify the frame length of 74 bytes

Foxiks:
It seems like it's ready now!

daniestevez:
Great. Squashing and merging.

PR #680, opened 2024-11-04
doppler: correct error when opening file failed

marcusmueller:
When using gr-satellites from an external project, it was a bit confusing to see format errors when they were actually just errors of files not being found.


daniestevez:
Thanks! I had noticed this before, but never bothered to fix it.

PR #689, opened 2024-11-30
Add check for digipeater callsign and hbit to block 'Check address'

DL7NDR:
Adds check for digipeater callsign and hbit set.
In combination with 'telemetry forwarder' good to filter out decoded own frames during transmitting (near TCA).

### DIFF
--- a/grc/satellites_check_address.block.yml
+++ b/grc/satellites_check_address.block.yml
@@ -33,4 +33,16 @@ templates:
     imports: import satellites
     make: satellites.check_address(${address}, ${direction}, ${digicallsign})
 
+documentation: |-
+    Enter an address (choose "from" for source or "to" for destination callsign) and/or enter a digipeater callsign.
+    The address callsign may have a SSID, the digipeater callsign must not have a SSID.
+    
+    Example and results:
+    Address: PCSAT-1 (from)
+    Digicallsign: W3ADO
+
+    Every frame with PCSAT-1 as source callsign will be forwarded to output port "ok".
+    Every frame with "via" W3ADO (SSID is arbitrary) and hbit set (usually marked with an asterisk (*) to show that it was digipeated) will be forwarded to output port "ok".
+    Everything else will be forwarded to output port "fail".
+    
 file_format: 1
--- a/grc/satellites_check_address.block.yml
+++ b/grc/satellites_check_address.block.yml
@@ -35,7 +35,9 @@ templates:
     make: satellites.check_address(${address}, ${direction}, ${digicallsign})
 
 documentation: |-
-    Enter an address (choose "from" for source or "to" for destination callsign) and/or enter a digipeater callsign.
+    Enter either an address (choose "from" for source or "to" for destination callsign) or a digipeater callsign or both.
+    The logical connective between 'address' and 'digipeater callsign' is an inclusive disjunction (OR).
+    If at least one callsign of both matches, the frame is passed to output port "ok".
     The address callsign may have a SSID, the digipeater callsign must not have a SSID.
     
     Example and results:
--- a/grc/satellites_check_address.block.yml
+++ b/grc/satellites_check_address.block.yml
@@ -13,6 +13,9 @@ parameters:
     dtype: string
     options: ['"from"', '"to"']
     option_labels: [From, To]
+-   id: digicallsign
+    label: Digicallsign
+    dtype: string
 
 inputs:
 -   domain: message
@@ -28,6 +31,6 @@ outputs:
 
 templates:
     imports: import satellites
-    make: satellites.check_address(${address}, ${direction})
+    make: satellites.check_address(${address}, ${direction}, ${digicallsign})
 
 file_format: 1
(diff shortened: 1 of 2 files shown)


daniestevez:
First off, it is not necessary to open and close multiple issues form the same thing (#684, #685, #686, #687, #688, #689). For each of these people subscribed to the repository notifications receive an email. You can push additional commits to the same branch or do `git commit --amend` and `git push --force`.

python/check_address.py
```diff
@@ -52,14 +53,42 @@ def handle_msg(self, msg_pmt):
         else:
             address = packet[7:14]
 
+        if len(packet) > 20:
```
daniestevez, line 56: What happens if `len(packet) <= 20`? It looks like `hbit` and `digi` are not defined, but they are read afterwards regardless.

python/check_address.py
```diff
@@ -52,14 +53,42 @@ def handle_msg(self, msg_pmt):
...
+# digi callsign; and above: hbit  (most significant bit after digi callsign)
+# (set to 1 in case of digipeated message)
+            digi = [c >> 1 for c in packet[14:20]]
+
+# error handling
+            try:
+                digi = bytes(digi).decode('ascii').rstrip(' ')
+
+            except UnicodeDecodeError:
+                print("Not to ASCII convertable string detected.")
```
daniestevez, line 68: How can this possibly fail to decode as ASCII? It is a `bytes` object where all the bytes have the MSB equal to zero. Regardless, in the event there was a `UnicodeDecodeError` exception, `digi` would be undefined but potentially read below.

python/check_address.py
```diff
@@ -52,14 +53,42 @@ def handle_msg(self, msg_pmt):
...
-        if (callsign != self.callsign
-                or (self.ssid is not None and ssid != self.ssid)):
-            # Incorrect address
-            self.message_port_pub(pmt.intern('fail'), msg_pmt)
-        else:
-            # Correct address
+
+# altered if sentence
+        if ((callsign == self.callsign
+                and (ssid == self.ssid if self.ssid is not None else True))
```
daniestevez, line 88: This condition is better written as `self.ssid is None or ssid == self.ssid`.

daniestevez (changes requested):
This feature should be optional and default to the current behaviour. The changes cannot be merged in their present form, because they change this block's behaviour and introduce a new constructor parameter without a default value, which breaks API compatibility.

python/check_address.py
DL7NDR: What happens if `len(packet) <= 20`?
->`hbit` and `digi` won't be defined.

Is it ok for you to initialize `hbit` and `digi` right before `len(packet) <= 20` by:
```
        hbit = 0
        digi = ""
```

daniestevez: I guess not. What's the code supposed to do in this case? If repeater address checking has been enabled but the message doesn't even contain repeater addresses because it is too short, shouldn't it go directly to the `fail` output?

python/check_address.py
DL7NDR: You are right. Then we just remove it.

python/check_address.py
DL7NDR: If `len(packet) <=20`, then hbit is 0 and there is no digipeater (digi = "").
In this case, the regular sender / destination callsign check should and will be processed anyway.

daniestevez: I had to read again the condition you wrote on lines 87-89 to understand what this feature is supposed to do. Let me rephrase: this feature will pass the frame to `ok` if either its destination address matches `address` or if the frame contains 3 addresses, the H bit is set (indicating that the frames has been repeated) and the repeater address (3rd address) matches `digicallsign`.

In that case I agree that just setting `hbit = 0` if `len(packet) <= 20` might be enough.

DL7NDR: > (...) if either its destination address matches `address` (...)
yes, destination or source (according to the chosen _direction_ "to" or "from")

Correct, that's the way I wanted it to work.

daniestevez:
@DL7NDR please take note of my previous remark:

> The changes cannot be merged in their present form, because they change this block's behaviour and introduce a new constructor parameter without a default value, which breaks API compatibility.

More concretely, the `digicallsign` argument in the block's class constructor should have a default value (possibly `''` or `None`), the GRC YAML file should also default to that value, and the default value should achieve the same behaviour that the block currently has (i.e., only the source or destination callsign is checked).

grc/satellites_check_address.block.yml
```diff
@@ -28,6 +32,18 @@ outputs:
 
 templates:
     imports: import satellites
-    make: satellites.check_address(${address}, ${direction})
+    make: satellites.check_address(${address}, ${direction}, ${digicallsign})
+
+documentation: |-
+    Enter an address (choose "from" for source or "to" for destination callsign) and/or enter a digipeater callsign.
+    The address callsign may have a SSID, the digipeater callsign must not have a SSID.
```
daniestevez, line 39: Can you describe accurately which frames are passed to the `ok` output? At the moment nothing in this documentation gives a clue as to whether the `or` in line 79 is an `or` or an `and`.

daniestevez (approved):
This looks good now. Approved and merging.

daniestevez:
I merged too fast. I didn't realize this was actually 7 commits. I should have squashed. I'll do this manually now.

PR #691, opened 2024-12-06
Add block 'Check ASCII address'

DL7NDR:
Checks data from input port "in" from the first byte on for an ASCII string entered in field "Address".
    
    If the incoming string matches the entered string, data will be passed to output port "ok", else to output port "fail".
    
    Unlike block 'check_address' (label: Check AX.25 address) 'check_address2' does not bitshift the bytes.
    If you are looking for the ASCII letter "A" (entered in field "Address"), the match would be 0x41. 'check_address' would match on 0x82.


daniestevez:
If I understood correctly, this block takes a string provided by the user, takes the first bytes of the PDU (as determined by the length of that string), attempts to convert these bytes to an ASCII string, and if the conversion is successful, it compares it against the user-provided string.

This is one small step of becoming something more general: take a list of bytes (`List[int]`) from the user and compare those with the first bytes of the PDU. Can you make these changes?

Since this doesn't have much to do with AX.25 addresses, another name for the block would be more suitable. Perhaps "Check prefix" (I haven't thought very hard about the name). And definitely avoid using `check_address2` as a class name.

Also, CI is failing.

DL7NDR:
Do you mean like this?

#!/usr/bin/env python3
# -*- coding: utf-8 -*-

# Copyright 2017 Daniel Estevez <daniel@destevez.net>
#
# This file is part of gr-satellites
#
# SPDX-License-Identifier: GPL-3.0-or-later
#

import struct

from gnuradio import gr
from typing import List
import pmt


class check_whatever(gr.basic_block):
    """docstring for block check_whatever"""
    def __init__(self, address: List[int]):
        gr.basic_block.__init__(
            self,
            name='check_whatever',
            in_sig=[],
            out_sig=[])

        self.callsign = bytes(address)

        self.message_port_register_in(pmt.intern('in'))
        self.set_msg_handler(pmt.intern('in'), self.handle_msg)
        self.message_port_register_out(pmt.intern('ok'))
        self.message_port_register_out(pmt.intern('fail'))

    def handle_msg(self, msg_pmt):
        msg = pmt.cdr(msg_pmt)
        if not pmt.is_u8vector(msg):
            print('[ERROR] Received invalid message type. Expected u8vector')
            return
        packet = bytes(pmt.u8vector_elements(msg))

        # Check packet length
        if len(packet) < len(self.callsign):
            self.message_port_pub(pmt.intern('fail'), msg_pmt)
            return

        callsign = packet[:len(self.callsign)]

        if (callsign == self.callsign):
            # match
            self.message_port_pub(pmt.intern('ok'), msg_pmt)
        else:
            # no match
            self.message_port_pub(pmt.intern('fail'), msg_pmt)

daniestevez:
Yes. Something along those lines.

DL7NDR:
Going to create a new merge request and closing this one.

PR #693, opened 2024-12-16
Update ASRTU-1.yml

janvgils:
Updating with the definite Norad ID.
The decoder was tested with the following observation https://network.satnogs.org/observations/10774576/


daniestevez (approved):
CI failure on this one is unrelated. It seems to be caused by an update of the `clang-format` version run by CI. I will fix this in another PR. Changes look good. Approved.

janvgils:
I will wait with the other requests until this is fixed.

PR #710, opened 2024-12-27
Add block 'Check Hex String'

DL7NDR:
Checks data from input port from "Starting Position" on for "Hex String".

    If the incoming string matches the entered string, data will be passed to output port "ok", else to "fail".

    Example:
    Hex String = c0ffee
    Starting Position = 1

    => Match for "ff"


DL7NDR (commented):
deleted "check_address2" files and reduced length of a line.

grc/satellites_check_hex_string.block.yml
```diff
@@ -0,0 +1,46 @@
...
+documentation: |-
+    Checks data from input port from "Starting Position" on for a "Hex String".
+
+    If the incoming string matches the entered string, data will be passed to output port "ok", else to "fail".
+
+    Example:
+    Hex String = c0ffee
+    Starting Position = 1
+
+    => Match for "ff"
```
daniestevez, line 43: Can you clarify the example? It is not clear which are the parameters for the block and which is the input PDU.

python/check_hex_string.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+# Copyright 2017 Daniel Estevez <daniel@destevez.net>
```
daniestevez, line 4: The copyright statement needs to be updated to the year and author.

python/check_hex_string.py
```diff
@@ -0,0 +1,52 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+
+
+from gnuradio import gr
+import pmt
+
+
+class check_hex_string(gr.basic_block):
+    """docstring for block check_hex_string"""
```
daniestevez, line 17: To be filled.

python/check_hex_string.py
```diff
@@ -0,0 +1,52 @@
...
+        msg = pmt.cdr(msg_pmt)
+        if not pmt.is_u8vector(msg):
+            print('[ERROR] Received invalid message type. Expected u8vector')
+            return
+        packet = bytes(pmt.u8vector_elements(msg))
+
+        # Check packet length
+        if len(packet) < len(self.hexstring):
+            self.message_port_pub(pmt.intern('fail'), msg_pmt)
+            return
```
daniestevez, line 43: This check should also take into account `self.startindex`.

python/check_hex_string.py
```diff
@@ -0,0 +1,52 @@
...
+
+        hex_str = packet[self.startindex:self.startindex + len(self.hexstring)]
+
+        if (hex_str == self.hexstring):
```
daniestevez, line 47: Parenthesis are not needed here.

daniestevez (changes requested):
I would have preferred that you had used #691 rather than closing it and opening a new PR, since the previous PR contains the discussion that impacted the design of these changes. But anyhow.

The code looks good except for a few comments that I've made. Can you also squash all the commits into a single one with a descriptive commit message? There is no reason to submit a PR with 21 commits just to add a simple block.

python/check_hex_string.py
DL7NDR: I've tested it with a startindex greater than the length of the string at input port. In this case it behaves as expected, the string from port "in" will be passed to port "fail". So, I would leave the startindex unchecked, if ok with you.

python/check_hex_string.py
DL7NDR: done.

python/check_hex_string.py
DL7NDR: done.

grc/satellites_check_hex_string.block.yml
DL7NDR: done.

python/check_hex_string.py
DL7NDR: Parenthesis removed.

python/check_hex_string.py
daniestevez: The potential problem is that in the line
```python
hex_str = packet[self.startindex:self.startindex + len(self.hexstring)]
```
below, if `len(packet) < self.startindex + len(self.hexstring)`, then you're slicing outside of the `packet` length. In Python this works, but you'll get a `hex_str` of the wrong length (shorter, potentially empty). This won't pass the equality match against `self.hexstring`. So either you can remove the check in L40-L43 altogether because Python's slicing takes care of failing the match if the packet is too short, or correct the L40-L43 packet length check to account for `self.startindex`.

DL7NDR:
> Can you also squash all the commits into a single one with a descriptive commit message?

Cannot find a function for that for github website only user.

daniestevez:
See for instance https://www.designgurus.io/answers/detail/how-do-i-squash-my-last-n-commits-together

python/check_hex_string.py
DL7NDR: decided to remove the check in L40-L43

DL7NDR:
> See for instance https://www.designgurus.io/answers/detail/how-do-i-squash-my-last-n-commits-together

I'm a github website -only- user.

daniestevez:
While the Github web UI is useful, it has limited functionality, and many operations need to be done with the `git` command line application.

> I'm a github website -only- user.

This comes across as "I don't want to learn how to use `git`, so others should do the work for me."

DL7NDR:
You are totally right, I don't want to learn `git`.
I just want to extend gr-satellites!
And if I really wanted to let others do the work for me, I wouldn't invest time in extending blocks by myself.

Not with bad intentions, but one might claim that also your sentence

> "Can you also squash all the commits (...)"

comes across as "_I want the commits to be squashed, but expect others to do the work for me._"

However, if you still insist on squashed commits, I'll give it another try. Just let me know.

daniestevez:
> comes across as "I want the commits to be squashed, but expect others to do the work for me."

On the contrary. Having a single commit for each set of changes that can be grouped as one unit is often regarded as good practice (and this means one commit per PR in most cases). gr-satellites (or GNU Radio) don't mention this explicitly in the contributing guidelines, but they stress that [commit messages are very important](https://github.com/daniestevez/gr-satellites/blob/main/CONTRIBUTING.md#git-commit-messages-are-very-important). Having 22 different commits with trivial changes just to add a new block goes against this recommendation. The duty to reformat git history appropriately falls on the PR submitter, not on the project maintainers.

> You are totally right, I don't want to learn git. I just want to extend gr-satellites!

Having to learn some git to contribute to a project that uses git as its VCS is a rather reasonable thing.

DL7NDR:
I'll try it with a new pull request to get rid of the unnecessary commits.

However, I'm interpreting "[commit messages are very important](https://github.com/daniestevez/gr-satellites/blob/main/CONTRIBUTING.md#git-commit-messages-are-very-important)" exactly the other way round.
Since by squashing commits the messages of every single commit are no longer directly visible (as far as I understood that), commits shouldn't be squashed to **keep** every commit message (because: "very important").

PR #712, opened 2025-01-04
Add these 5 files for block check_hex_string

DL7NDR:
Checks data from input port from "Starting Position" on for a "Hex String".

    If the incoming string matches the entered string, data will be passed to output port "ok", else to "fail".

    Example for a match:

    Entered block parameters:
    "Hex String" = 4450304E4641	(ASCII for DP0NFA)
    "Starting Position" = 10		(Starts counting from 0 on)

    Data coming into input port: "02300400000000003D0E4450304E4641000100C300B7142EF90F9C06290000000"


DL7NDR:
Instead of editing files and adding files individually, I just dragged and dropped the folders grc and python with the 5 files. This created only one commit, as requested.

daniestevez:
That worked, although the commit title "Add these 5 files for block check_hex_string" was a bit awkward and I modified it to "Add Check String Block". But next time that you're contributing to any project that uses git as VCS, do yourself and the maintainers a favour and learn some `git` command line. The Github and Gitlab web UIs are fine for a limited number of operations, but most of the required operations (even the ones needed on a daily basis) cannot be done on the web UI.

N6RFM:
Thank you Daniel and Dani!  

Very much looking forward to trying this new block. While the Check AX.25 Address is sometimes helpful, this should provide a nice and complimentary filtering tool once a consistent hex string for a given satellite is identified as many satellites do not transmit To or From call signs.   TKS AGN es 73, Bob N6RFM

PR #713, opened 2025-01-20
Create COLIBRI-S.yml

janvgils:
The Yaml file is made based on an IQ recording and measured in the following way.

`gr_satellites COLIBRI-S.yml --rawint16 cut_tuned_iq_57600_10956048.wav --samp_rate 57600 --iq --disable_dc_block --dump_path .`

That produced the following output.

image


daniestevez:
Looks good. Merging.

Issue #720, opened 2025-02-20
Add basic support for SigMF input


kerel-fs:
IQ recordings can only be understood when having a minimum of metadata, namely the datatype of the stored samples and the sample rate. Currently this metadata can be passed to gr-satellites via CLI options (`--rawfile` or `--rawint16`,  `--iq` and `--samp_rate`).

The [SigMF](https://sigmf.org) specification describes how to store this metadata alongside a recording, enabling applications to understand compliant recordings without user-provided metadata.

💡 **Proposed minimal solution**

- Add new option `--sigmf_in SIGMF_IN` that can be used to read a SigMF recording. The file is interpreted to be the SigMF Metadata file. Read the datatype and sample rate from this metadata file and set the input source accordingly.

🚀 **Possible future enhancements**

- SigMF Archives
  The latest SigMF specification describes SigMF Archive files, which are `tar` archives, that contain one or more SigMF Recording.
  
  An advantage of using a SigMF Archive file is the fact that it is a single file, whereas a SigMF recording always consist of at least two files: the SigMF Metadata file and the SigMF Dataset file.
  A current disadvantage of using a SigMG Archive file is the current lack of support by most tools.

---
Example usage:
```
gr_satellites "Reaktor Hello World" --sigmf_in /satnogs_data/rf/iq/pi9cam/iq_1452119.sigmf-meta
```

daniestevez:
Thanks for writing a proposal as an issue and also submitting a PR!

Some comments about the design.

1. Could we call the parameter `--sigmf` rather than `--sigmf_in`? This is a minor nit, but it feels more consistent with other parameters (e.g., `--wavfile` and `--rawfile` rather than `--wavfile_in` and `--rawfile_in`).

2. Could we allow that the value of the `--sigmf` parameter is either the `.sigmf-data` file or the `.sigmf-meta` file? This is reasonably easy to handle in code, and it saves the user from having to remember which of them should be passed. Other tools do it like so. For instance, you can either run `inspectrum file.sigmf-data` or `inspectrum.sigmf-meta`. Moreover, if/when SigMF archive support is implemented, this could also be done by detecting whether the extension of the value passed is `.sigmf`.

[...]

PR #721, opened 2025-02-20
Add basic support for SigMF input

kerel-fs:
Follows the proposal from #720.

Currently supported & tested data types:
- `ci16_le`

This list can probably be extended quite easily, for the first iteration I wanted to focus on what I can test.

### DIFF
--- a/apps/gr_satellites
+++ b/apps/gr_satellites
@@ -9,6 +9,7 @@
 #
 
 import argparse
+import json
 import signal
 import sys
 
@@ -86,6 +87,7 @@ def argument_parser():
     p_sources.add_argument(
         '--udp', action='store_true', help='Use UDP input')
     p_sources.add_argument('--kiss_in', help='KISS input file')
+    p_sources.add_argument('--sigmf', help='SigMF input file')
 
     p_input.add_argument('--samp_rate', type=float, help='Sample rate (Hz)')
     p_input.add_argument(
@@ -133,9 +135,10 @@ def argument_parser():
 def check_options(options, parser):
     if (options.kiss_in is None
             and options.wavfile is None
-            and options.samp_rate is None):
+            and options.samp_rate is None
+            and options.sigmf is None):
         print('Need to specify --samp_rate unless --wavfile '
-              'or --kiss_in is used',
+              'or --kiss_in or --sigmf is used',
               file=sys.stderr)
         parser.print_usage(file=sys.stderr)
         sys.exit(1)
@@ -227,6 +230,8 @@ class gr_satellites_top_block(gr.top_block):
             return self.setup_udp_input()
         elif self.options.kiss_in is not None:
             return self.setup_kiss_input()
+        elif self.options.sigmf is not None:
+            return self.setup_sigmf_input()
         else:
             raise Exception('No input source set for flowgraph')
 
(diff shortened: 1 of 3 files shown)


apps/gr_satellites
```diff
@@ -292,6 +297,61 @@ class gr_satellites_top_block(gr.top_block):
...
 
+    def setup_sigmf_input(self):
+        if self.options.sigmf_in.endswith('sigmf'):
+            print(
+                "Failed to load SigMF archive (archives are not yet supported). "
+                "Consider extracting a recording and pass the corresponding SigMF metadata file.",
+                file=sys.stderr)
+            sys.exit(1)
+        if not self.options.sigmf_in.endswith('sigmf-meta'):
+            print("SIGMF_IN has invalid file extension, expected '.sigmf-meta'", file=sys.stderr)
```
daniestevez, line 308: Maybe we can print the actual `self.options.sigmf_in` that the user passed instead of the placeholder `SIGMF_IN`. That would make it even more obvious for the user what is wrong with the CLI arguments.

apps/gr_satellites
```diff
@@ -292,6 +297,61 @@ class gr_satellites_top_block(gr.top_block):
     def setup_kiss_input(self):
         self.input = datasources.kiss_file_source(self.options.kiss_in)
 
+    def setup_sigmf_input(self):
+        if self.options.sigmf_in.endswith('sigmf'):
+            print(
+                "Failed to load SigMF archive (archives are not yet supported). "
```
daniestevez, line 303: There is an inconsistent usage of double quotes (`"`) and simple quotes (`'`). In this PR. I believe that the gr-satellites code base only uses simple quotes to denote strings.

apps/gr_satellites
```diff
@@ -292,6 +297,61 @@ class gr_satellites_top_block(gr.top_block):
...
+            sys.exit(1)
+
+        fields_required = ['core:datatype', 'core:sample_rate']
+        for field in fields_required:
+            if field not in meta['global']:
+                print(f"SigMF metadata file is malformed: Missing field '{field}'", file=sys.stderr)
+                sys.exit(1)
+
+        datatype = meta['global']['core:datatype']
+        dataset_filename = self.options.sigmf_in[:-11] + '.sigmf-data'
```
daniestevez, line 335: Maybe this, as it makes it more obvious where the number `11` comes from:
```suggestion
        dataset_filename = self.options.sigmf_in[:-len('.sigmf-meta')] + '.sigmf-data'
```

apps/gr_satellites
```diff
@@ -292,6 +297,61 @@ class gr_satellites_top_block(gr.top_block):
...
+        else:
+            print(f"Unsupported SigMF Dataset format of stored samples ('{datatype}'). Supported datatypes are: {', '.join(datatypes_supported)}.", file=sys.stderr)
+            sys.exit(1)
+
+        if 'captures' in meta and 'core:datetime' in meta['captures']:
+            self.options.start_time = meta['captures']['core_datetime']
+
+        if 'capture' in meta and 'core:sample_start' in meta['captures'] and meta['captures']['core:sample_start'] != 0:
+            print('SigMF Capture Segments with non-zero sample start index are not supported (yet).', file=sys.stderr)
+            sys.exit(1)
```
daniestevez, line 354: Some comments about captures.

1.  `if 'capture' in meta` and `'core_datetime'` look like typos.

2. [According to the SigMF spec](https://sigmf.org/#sigmf-metadata-format), "The top-level Object MUST contain three JSON Objects named `global`, `captures`, and `annotations`.", so we should fail and exit if `'captures' not in meta`.

3. `meta['captures']` is going to be a `list` (even if there is a single capture). I would probably only set `self.options.start_time` if there is a single capture that contains `core:datetime` and has `core:sample_start` equal to zero. I would allow other types of captures (as long as they are not malformed SigMF metadata), but not set the `self.options.start_time` in that case (and perhaps, if there was a `core:datetime` somewhere, print a warning that this `datetime` was not applied). Being tolerant in what we accept might be useful, if for instance a user has a dataset that they want to process, but the dataset contains multiple captures with time discontinuities. The user might not care about setting `start_time`, so we should allow them to process this dataset.

4. [According to the SigMF spec](https://sigmf.org/#subsec:CapturesArray), `core:sample_start` is a required field in each capture, so probably we want to fail here if `core:sample_start` is not present.

kerel-fs: 1. typo fixed (good catch, apparently I never tested if the `start_time` is properly applied).
2. My original idea was to enforce only what is strictly necessary for gr-satellites, and be permissive regarding non-compliant metadata files when possible. However, by enforcing SigMF spec compliance the parser does become simpler. So, I've added a check to enforce `global`, `captures` and `annotations` as top-level objects. 
3. implemented as you described.
4. Done. Note that `[]` is also a valid value for the `captures` object. So either `captures` is `[]`, or we require `core:sample_start`.

kerel-fs:
Updated the PR addressing your comments, including the suggestions 1 & 2 from https://github.com/daniestevez/gr-satellites/issues/720#issuecomment-2673987186.

daniestevez (approved):
This looks great now. Approved and merging.

PR #725, opened 2025-03-04
crcs: Use digital.crc_check for GNU Radio >= 3.10

kerel-fs:
Pleasant side-effect: The upstreamed block is less verbose and does not print to stdout/stderr by default.

### DIFF
--- a/python/components/deframers/fossasat_deframer.py
+++ b/python/components/deframers/fossasat_deframer.py
@@ -15,7 +15,7 @@ import pmt
 from ... import pdu_head_tail
 from ... import sx12xx_packet_crop
 from ... import reflect_bytes
-from ... import crc_check
+from ...crcs import crc_check
 from ...hier.pn9_scrambler import pn9_scrambler
 from ...hier.sync_to_pdu_packed import sync_to_pdu_packed
 from ...utils.options_block import options_block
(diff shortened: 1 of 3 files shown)


daniestevez:
The Check CRC block appeared in GNU Radio 3.10.2.0, so the version check would need to be more complex than >= 3.10 if we want to support 3.10.0 and 3.10.1 (and probably we want, since we're also still supporting 3.8 and 3.9).

kerel-fs:
Updated to check against the minor version too.

In an earlier attempt to update this patch, I [added](https://github.com/kerel-fs/gr-satellites/blob/gnuradio_compat/python/utils/gnuradio_compat.py) a helper method `gnuradio_version_older_than(major: int, api: int, minor: int) -> bool` to make the check re-useable and to make it easy to identify locations where we check for old versions. This code is available in my branch [`gnuradio_compat`](https://github.com/kerel-fs/gr-satellites/commits/gnuradio_compat/) currently. I've since changed my mind about it's usefulness, thus it's not part of this patch. If you think different let me know so I can adapt it in this patch.

daniestevez:
From a grep I see that there is also

https://github.com/daniestevez/gr-satellites/blob/main/python/components/deframers/openlst_deframer.py#L14

kerel-fs:
It's included in my last commit already. I had grepped to find all such occurrences when preparing it.

PR #726, opened 2025-03-16
Update check_address.py

DL7NDR:
TEVEL2 sender callsigns (example: TVL2-3-1) using one "-" within the callsign showed that check_address.py splitted on every "-" instead on the last.

Changing line 27 from 'split' to "rsplit' solves the problem.


daniestevez:
Looks good. Merging.

PR #727, opened 2025-03-17
Switch to cmake 3.13

AndreiCherniaev:
stop to use make because since CMake 3.15 (Ubuntu 20.04), Generator can be set in env CMAKE_GENERATOR and sometimes it is not the Unix Makefiles; Tested in Ubuntu 24
P.S. Also you can install to non-system folder, see example https://askubuntu.com/a/1543849/1087530

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@
 ########################################################################
 # Project setup
 ########################################################################
-cmake_minimum_required(VERSION 3.13)
+cmake_minimum_required(VERSION 3.8)
 project(gr-satellites CXX C)
 enable_testing()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@
 ########################################################################
 # Project setup
 ########################################################################
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.13)
 project(gr-satellites CXX C)
 enable_testing()
 
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -95,11 +95,11 @@ inside the directory containing the gr-satellites sources:
 
 .. code-block:: console
 
-   $ mkdir build
-   $ cd build
-   $ cmake ..
-   $ make
-   $ sudo make install
+   $ git clone https://github.com/daniestevez/gr-satellites
+   $ cd gr-satellites/
+   $ cmake -S . -B build
+   $ cmake --build build --parallel
+   $ cmake --install build
    $ sudo ldconfig
 
 After running ``make``, you can run the tests by doing ``make test`` in the
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -95,14 +95,12 @@ inside the directory containing the gr-satellites sources:
 
 .. code-block:: console
 
-   $ git clone https://github.com/daniestevez/gr-satellites
-   $ cd gr-satellites/
    $ cmake -S . -B build
    $ cmake --build build --parallel
-   $ cmake --install build
+   $ sudo cmake --install build
    $ sudo ldconfig
 
-After running ``make``, you can run the tests by doing ``make test`` in the
+After build stage, you can run the tests by doing ``make test`` in the
 ``build/`` directory.
 
 .. note::


AndreiCherniaev:
> Can you clarify why CMake >= 3.13 is required?

In according with [link](https://cmake.org/cmake/help/latest/manual/cmake.1.html#generate-a-project-buildsystem),  `-B <path-to-build>` added in version 3.13.
P.S. And CMake 3.13 is very old software. For example [Qt project](https://github.com/qt/qtbase/blob/a2e60ebee3737548d1be14fdbb39b08c515ae602/CMakeLists.txt#L8) for now use 3.16

daniestevez:
There is nothing wrong with the instructions in the documentation. `mkdir build && cd build && cmake .. && make && sudo make install` is a perfectly fine way to install CMake projects. There are many other possible ways, and people that want to install to a specific directory or to do something else will know how to do it.

So if the whole point of requiring CMake >= 3.13 is to allow the use of `cmake -B` in documentation, to me that doesn't sound enough of a reason. To be honest, I don't know if there is anything in the CMakeLists.txt of gr-satellites that breaks with CMake 3.8, since this isn't tested in CI. But my point was about why to make this change now.

AndreiCherniaev:
> There is nothing wrong with the instructions in the documentation. `mkdir build && cd build && cmake .. && make && sudo make install` is a perfectly fine way to install CMake projects. There are many other possible ways, and people that want to install to a specific directory or to do something else will know how to do it.
> 
> So if the whole point of requiring CMake >= 3.13 is to allow the use of `cmake -B` in documentation, to me that doesn't sound enough of a reason. To be honest, I don't know if there is anything in the CMakeLists.txt of gr-satellites that breaks with CMake 3.8, since this isn't tested in CI. But my point was about why to make this change now.

Please don't directly call make in modern code. Since CMake 3.15 (Ubuntu 20.04), Generator can be set in env CMAKE_GENERATOR and sometimes it is not the Unix Makefiles... This I main reason. But there are others:
2) don't ask user to manage multi-threading building by himself, cmake can do it automatically by --parallel; 
3) use -B option to avoid `cd`. In this case user can more easy take your code and put it to complex script with a lot of things...

AndreiCherniaev:
> I'm yet to find a system

Try Ubuntu 24 or any, by default there is no `make` application, and you no need it to build gr-satellites or other software... Just for example if you try to build Qt with make you probably will have a error (because you need ninja).
For example this is all what you need in Ubuntu 24, one more time, I have no `make`:
```
sudo apt install g++ cmake ninja-build  git gnuradio-dev bzip2
cmake -S . -B build
cmake --build build --parallel
sudo cmake --install build
```
And of cause Windows and macOS hasn't `make`. I suggest way which is really nice. 99% users will do as your instruction says that is why for me so important to provide modern cmake code to allow people to have good experience. When I say modern I mean from 2018 year...

> you have failed to account for its context
> but that typically needs sudo
Thanks, fixed.

daniestevez:
> Try Ubuntu 24 or any, by default there is no make application

By default there is no `cmake` or `g++` either. That's the reason you are installing them. `make` is a dependency of `build-essential`.

> Just for example if you try to build Qt with make you probably will have a error (because you need ninja).

Why is this relevant? gr-satellites is not Qt.

> For example this is all what you need in Ubuntu 24 (one more time, I have no make in system):

Sure you can build gr-satellites however you like on your system, and you can avoid installing `make` if you want to. The point of this PR is not how one particular person wants to builds gr-satellites on their system. It is whether we change the build instructions in the documentation. These build instructions have worked fine for years and still work in most cases.

> And of cause Windows and macOS hasn't make.

Windows doesn't have anything of what is required to build GNU Radio out-of-tree modules (including, but not limited to, gr-satellites). This is true regardless of whether you build in the way you mention. It is quite cumbersome to get a suitable build environment in Windows, and these days probably the best way is through conda.

I don't know enough about macOS to tell if it has `make` by default, but I would find it surprising that installing `make` is any harder than installing `cmake` and `ninja`.

> 99% users will do as your instruction says that is why for me so important to provide modern cmake code to allow people to have good experience.

The current instructions will work fine for 99% of the users that are building from source.

AndreiCherniaev:
Still think gr-satellites should use modern cmake code.

PR #737, opened 2025-04-29
[maint-3.8] Add Mobitex-NX deframer

daniestevez:
Backport to `maint-3.8` of #723.


daniestevez:
I have fixed some errors in the QA tests to backport this, but I have arrived to a larger obstruction: the Matrix Interleaver block [was added in GNU Radio 3.8.4](https://github.com/gnuradio/gnuradio/releases/tag/v3.8.4.0), but Ubuntu 20.04, which is what the `maint-3.8` CI uses, [has GNU Radio 3.8.1](https://packages.ubuntu.com/focal/gnuradio).

It seems to me that backporting this new Mobitex-NX block to the `maint-3.8` branch is perhaps not enough reason to drop compatibility with GNU Radio 3.8.1, and that it isn't worth the effort to work on this backporting.

Issue #744, opened 2025-05-07
Mobitex: Program hangs after `tagged_stream_to_pdu :critical: Missing a required length tag`


kerel-fs:
A problem with the new builtin Mobitex deframer has been identified.

**Problem**

Running `gr_satellites TUBIN --rawint16 iq_9622036_48000.wav --samp_rate 48e3 --iq` successfully decodes some frames, then the following message is printed and the program hangs:
```
tagged_stream_to_pdu :critical: Missing a required length tag on port 0 at item #3600
thread_body_wrapper :error: ERROR thread[thread-per-block[12]: <block tagged_stream_to_pdu(27)>]: Missing length tag.
```

The `iq_9622036_48000.wav` IQ file can be found at https://www.pe0sat.vgnet.nl/download/TUBIN/

**Investigation**

The `tagged_stream_to_pdu` block is used to re-pack the stream of symbols into PDUs/messages, where each message contains one Mobitex block. Somewhere before that, a `packet_len` tag is lost, which is causing this error.
The flow described here is defined by: https://github.com/daniestevez/gr-satellites/blob/9687a727ffec2637fbabbc3229f7c7f23a084e36/python/components/deframers/mobitex_deframer.py#L146-L161

The `packet_len` tag is generated by the `pdu_to_tagged_stream` block, which consumes messages with individual Mobitex data blocks, and produces a stream of symbols with one `packet_len` tag per data block.

I'll continue this investigation myself soon, but since the code is part of the `main` branch now and thus a bug, I wanted to share this timely, even if the fix is not done yet.

kerel-fs:
- Testing with a bypassed scrambler still produces the error.
- Testing with a __bypassed matrix interleaver makes the error disappear__.

Thus, I conclude that the  `blocks.matrix_interleaver` has certainly something to do with it.

The exact timing of the error is not deterministic. Initially I assumed this might be caused by the input to the deframer being slightly different for each run (maybe caused by the FSK demodulator). To validate this assumption, I modified the code to dump the hard symbols right after the binary slicer to a file, and then replay this file and feed the `self.sync2pdu` block with it. Doing so, the  "Missing a required length tag" error could still be seen, and the timing remained non-deterministic. My code for this is in the [`mobitex_debug`](https://github.com/kerel-fs/gr-satellites/tree/mobitex_debug) branch. I though about adding a generic option to use a deframer with hard-symbols. But given the issue here would not benefit from it, I did not pursue this idea further yet.

[...]

PR #749, opened 2025-05-17
Mobitex deframer: Replace matrix interleaver with  block interleaver

kerel-fs:
Explanation (taken from [1]):
The matrix interleaver block does tag propagation badly. Tags get replicated many times in the output, which was causing issues. Moreover, the implementation of the matrix interleaver using Deinterleave + Interleave (with `nrow` connections in between) puts too much burden on the GNU Radio runtime for performing such a simple task.

This patch replaces the matrix interleaver with the block interleaver. The order of indices (`.reshape(12, 20)`) was confirmed by validating the output with this patch applied to the previous known-good output for a test recording from [2].

Fixes https://github.com/daniestevez/gr-satellites/issues/744.

[1]: https://github.com/daniestevez/gr-satellites/issues/744#issuecomment-2871281167
[2]: https://pe0sat.vgnet.nl/download/TUBIN/

### DIFF
--- a/python/components/deframers/mobitex_deframer.py
+++ b/python/components/deframers/mobitex_deframer.py
@@ -11,6 +11,7 @@
 
 import importlib
 import pmt
+import numpy as np
 
 from gnuradio import gr, blocks, digital
 from gnuradio.pdu import pdu_set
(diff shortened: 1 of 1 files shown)


daniestevez (approved):
This looks good to me. A disadvantage of this approach is that the Block Interleaver first appeared in GNU Radio 3.10.4 (that was ~3 years ago). I didn't realise this when proposing this solution. However, now that I know this, this still seems the appropriate solution to me. Unfortunately this means that we cannot backport this fix to the `maint-3.9` branch.

PR #751, opened 2025-05-17
pdu_to_kiss: Properly handle timezone-aware datetimes

### LINKED ISSUE #750: TypeError when using KISS output with sigmf input
**Problem**

When using a sigmf recording as input and a KISS output file as output, gr-satellites fails & hangs with the following error message:

```
thread_body_wrapper :error: ERROR thread[thread-per-block[17]: <block pdu_to_kiss(8)>]: TypeError: can't subtract offset-naive and offset-aware datetimes
At:
  /usr/local/lib/python3.13/site-packages/satellites/pdu_to_kiss.py(62): create_timestamp
  /usr/local/lib/python3.13/site-packages/satellites/pdu_to_kiss.py(79): handle_msg
```

This is caused by a mismatch of datetime types in the "PDU to KISS" block. The timestamp provided via sigmf metadata is timezone-aware, whereas the "PDU to KISS" block expects a timezone-naive time in `initial_timestamp`.

**System Information**

- gr-satellites version: c95343d1 (main)
- IQ files: https://pe0sat.vgnet.nl/download/TUBIN/

**Steps to reproduce**

1. Run gr-satellites with sigmf input and kiss file output:
```
gr_satellites TUBIN --sigmf iq_48000_10890339.sigmf-data --kiss_out out.kiss
```

**Possible solution**

In "PDU to KISS", convert `initial_timestamp` to UTC timezone, then drop the tzinfo to continue using timzone-naive timestamps internally. 

The "recommended" method by Python upstream is to use timezone-aware timestamp everywhere, but for well-controlled components like a GNU Radio block, I do not see much advantage in this (I'm indifferent on both options).

kerel-fs:
The output of 'parse_time' is expected to be timezone-naive. To achieve this, convert timezone-aware datetimes to UTC, then drop tzinfo.

Fixes #750

### DIFF
--- a/python/submit.py
+++ b/python/submit.py
@@ -20,7 +20,14 @@ import numpy
 
 def parse_time(time):
     try:
-        return datetime.datetime.fromisoformat(time)
+        time = datetime.datetime.fromisoformat(time)
+
+        if time.tzinfo:
+            # Ensure time is converted to UTC, then drop tzinfo
+            time = time.astimezone(datetime.timezone.utc)
+            time = time.replace(tzinfo=None)
+
+        return time
     except AttributeError:
         # Workaround for Python version <3.7, which doesn't have
         # fromisoformat()


daniestevez (approved):
The fix looks good to me. At some point we will need to refactor this block, because the whole reason that the return value of `parse_time()` needs to have no timezone is that the block uses `datetime.datetime.utcnow()` to get the current time, but that function is deprecated since Python 3.12, and `datetime.datetime.now(datetime.timezone.utc)` should be used instead.

Issue #752, opened 2025-05-18
pdu_to_kiss: refactor timezone handling


daniestevez:
The `pdu_to_kiss` block uses `datetime.datetime.utcnow()`, which is deprecated since Python 3.12. `datetime.datetime.now(datetime.timezone.utc)` should be used instead. Moreover, see #750 for the possible pitfalls of using timezones in Python. The block needs to be refactored to handle timezones properly, but compatibility with Python 3.8 needs to be maintained.

PR #769, opened 2025-10-01
pdu_to_kiss: Use dateutil.parser if python <3.11 and available

### LINKED ISSUE #767: SigMF input: Datetime parser fails for valid input (Python <=3.10)
**Description**
gr-satellites is using Pythons [datetime.fromisoformat(date_string)](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat) method to parse the timestamp provided by `--start_time` or via the SigMF metadata file (see [python/submit.py#L23](https://github.com/daniestevez/gr-satellites/blob/74c3d604c4ed010101a1b653e0499ee2129f7be8/python/submit.py#L23)).

In Python <=3.10, this method does not support all valid ISO 8601 formats, including the subset of formats specified by SigMF.

When decoding from a valid SigMF file, gr-satellites thus fails with a traceback.

📜 **Error log**

(This log is fabricated. I personally don't have a host with Python 3.10 and gr-satellites atm):
```
❯ gr_satellites "Reaktor Hello World" --sigmf iq_1452119.sigmf-meta
Traceback (most recent call last):
  File "/usr/local/bin/gr_satellites", line 410, in <module>
    main()
    ~~~~^^
  File "/usr/local/bin/gr_satellites", line 395, in main
    tb = gr_satellites_top_block(parser)
  File "/usr/local/bin/gr_satellites", line 202, in __init__
    self.flowgraph = satellites.core.gr_satellites_flowgraph(
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        samp_rate=samp_rate, iq=self.options.iq, pdu_in=pdu_in,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        options=options, config=self.config, dump_path=options.dump_path,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        **sat)
        ^^^^^^
  File "/usr/local/lib/python3.10/site-packages/satellites/core/gr_satellites_flowgraph.py", line 147, in __init__
    self._init_additional_datasinks()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.10/site-packages/satellites/core/gr_satellites_flowgraph.py", line 222, in _init_additional_datasinks
    self.get_telemetry_submitters(self.satyaml, self.config,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                  self.options))
                                  ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.10/site-packages/satellites/core/gr_satellites_flowgraph.py", line 343, in get_telemetry_submitters
    datasinks.telemetry_submit(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^
        'SatNOGS', norad=norad, config=config, options=options)]
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.10/site-packages/satellites/components/datasinks/telemetry_submit.py", line 50, in __init__
    self.submit = submit(url, norad,
                  ~~~~~~^^^^^^^^^^^^
                         config['Groundstation']['callsign'],
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                         float(config['Groundstation']['longitude']),
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                         float(config['Groundstation']['latitude']),
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                         initial_timestamp)
                         ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.10/site-packages/satellites/submit.py", line 60, in __init__
    parse_time(initialTimestamp)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.10/site-packages/satellites/submit.py", line 24, in parse_time
    time = datetime.datetime.fromisoformat(time)
ValueError: Invalid isoformat string: '2019-12-28T20:03:28Z'
```

Minimal reproducer:
```
❯ python
Python 3.10.18 (main, Aug  8 2025, 17:07:22) [Clang 20.1.4 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from datetime import datetime
>>> datetime.fromisoformat('2019-12-28T20:03:28Z')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Invalid isoformat string: '2019-12-28T20:03:28Z'
```

This was reported to me by @DL4PD and originally I considered this a bug, but upon further inspection this seems to be fixed for Python 3.11+. Before I start and prepare a patch, I'd like to know:

Which Python versions does gr-satellites support currently?

kerel-fs:
Proposed Workaround for missing ISO-8601 timestamp format support in older Python versions.

Fixes #767.

Tested with Python 3.13.5 only (aka "no workaround used").

**Stand-alone test script**

To test this code with different Python versions, I've used the following stand-alone script: 
[issue767-parse_time.py](https://github.com/user-attachments/files/22649850/issue767-parse_time.py).

Logs:
```
❯ uv venv --python 3.8
Using CPython 3.8.20
Creating virtual environment at: .venv
Activate with: source .venv/bin/activate
❯ source .venv/bin/activate
❯ ./issue767-parse_time.py
Failed to parse start timestamp, Error:
Invalid isoformat string: '2025-05-01T01:01:12.123400Z'

Recommendation (with Python Version <= 3.10):
Install package 'python-dateutil' for improved timestamp parser support.
❯ uv pip install python-dateutil
Resolved 2 packages in 5ms
Installed 2 packages in 6ms
 + python-dateutil==2.9.0.post0
 + six==1.17.0
❯ ./issue767-parse_time.py
2025-05-01 01:01:12.123400
```

### DIFF
--- a/python/telemetry/cirbe_70cm.py
+++ b/python/telemetry/cirbe_70cm.py
@@ -43,16 +43,11 @@ class TimeAdapter(Adapter):
         return Container()
 
     def _decode(self, obj, context, path=None):
-        t = datetime.datetime.fromtimestamp(
-            obj.time_stamp_seconds,
-            tz=datetime.timezone.utc
+        j2000_epoch = datetime.datetime(2000, 1, 1, 12,
+                                        tzinfo=datetime.timezone.utc)
+        return j2000_epoch + datetime.timedelta(
+            seconds=obj.time_stamp_seconds
         )
-        t = t.replace(tzinfo=None)
-        offset = (
-            datetime.datetime(2000, 1, 1, 12)
-            - datetime.datetime(1970, 1, 1)
-        )
-        return (t + offset)
 
 
 SecondaryHeader = TimeAdapter(


daniestevez (commented):
Wow. This is really complicated. It has a bunch of different use cases, it tries to use an external package that may or may not be installed and has fallback cases. I think this would work correctly in all cases, but I'm not sure if it is a good idea to introduce this complexity.

This change is related to Python `datetime` handling in general, including naïve versus timezone aware objects. This has to do with https://github.com/daniestevez/gr-satellites/issues/752. At some point we need to remove uses of `utcnow()`, which this block also uses.

So probably it is good to address all the `datetime` quirks as a whole.

My impression is that if the issue with Python 3.10 is that it doesn't like the `Z` at the end of the SigMF timestamps, then we could `.rstrip('Z')` the SigMF timestamp. That is ugly, but would work (note that then we get a naïve object, which might not be what we want). I certainly prefer that to a 60 line function to just parse a timestamp.

I'm out of bandwidth at the moment, but I might investigate more about what works in different Python versions and come up with a better solution, specially since #752 has been opened for quite a while and at some point that will come to bite us.

daniestevez:
I've done some testing and documentation reading and have come up with a concept of how `datetime` should be handled in gr-satellites.

This is illustrated in the following test script `timestamp_test.py`:

```python
#!/usr/bin/env python3

import datetime
import sys


def parse_timestamp(s):
    if sys.version_info < (3, 11):
        # in Python 3.10, fromisoformat() does not accept
        # a trailing Z, and it can only parse millisecond
        # resolution. We remove any trailing Z, and parse
        # the fractional seconds separately
        s = s.rstrip('Z')
        s_split = s.split('.')
        s_int, s_frac = s.split('.')
        t_int = datetime.datetime.fromisoformat(s_int)
        t_frac = float(f'0.{s_frac}')
        # Truncate t_frac to integer microseconds to match
        # behaviour of fromisoformat() in newer Python version
        t = t_int + datetime.timedelta(
            microseconds=int(t_frac * 1000000))
    else:
        t = datetime.datetime.fromisoformat(s)
    t = t.replace(tzinfo=datetime.timezone.utc)
    return t


def main():
    # parse timestamps in various formats
    t0 = parse_timestamp('2025-10-02T17:49:22.123456789Z')
    assert str(t0) == '2025-10-02 17:49:22.123456+00:00'

    t1 = parse_timestamp('2025-10-02T20:47:24.587123442')
    assert str(t1) == '2025-10-02 20:47:24.587123+00:00'

    t2 = parse_timestamp('2025-10-02T20:50:03.157291322Z')
    assert str(t2) == '2025-10-02 20:50:03.157291+00:00'

    # get current time
    t_now = datetime.datetime.now(tz=datetime.timezone.utc)
    print(f'now = {t_now}')

    # get elapsed time
    elapsed = (t1 - t0).total_seconds()
    assert elapsed == 3 * 3600 - 2 * 60 + 2 + 0.587123 - 0.123456

    # propagate time (as in submit.py)
    t_prop = t2 - t1 + t0
    assert str(t_prop) == '2025-10-02 17:52:00.693624+00:00'

    # format as needed in submit.py
    t0_fmt = t0.replace(tzinfo=None).isoformat()[:-3] + 'Z'
    assert t0_fmt == '2025-10-02T17:49:22.123Z'

    # compute KISS timestamp
    t0_kiss = round(t0.timestamp() * 1e3)
    assert t0_kiss == 1759427362123


if __name__ == '__main__':
    main()
```

I'm not too proud of how parsing for Python <= 3.11 turned out, but this works fine for Python >= 3.8, <=3.13:
```
$ for n in $(seq 8 13); do uv run --python 3.${n} timestamp_test.py; done
now = 2025-10-02 21:06:17.445684+00:00
now = 2025-10-02 21:06:17.459511+00:00
now = 2025-10-02 21:06:17.473154+00:00
now = 2025-10-02 21:06:17.486486+00:00
now = 2025-10-02 21:06:17.498104+00:00
now = 2025-10-02 21:06:17.511073+00:00
```

So I think we need to modify `datetime` usages in `pdu_to_kiss`, `submit` and elsewhere to match this example (a grep for `datetime` inside the `python` folder shows other usages).

kerel-fs:
Agreed, I did not like the complexity my proposed workaround would have introduced either. It was inspired by the Python documentation referencing `python-dateutil`, specifically [3.10: datetime.fromisoformat](https://docs.python.org/3.10/library/datetime.html#datetime.datetime.fromisoformat).

I welcome that we can drop Python <=3.7, and especially like that this manual parser is matching the microsecond behaviour of Python >=3.11. This was the missing piece from my own (similar) attempts at a manual parser that made me search for an alternative.

The concept looks good. I've implemented your proposal verbatim, and then took one pass searching for `datetime` to replace all usage of the deprecated methods with the given alternatives. I've decided to not alter any API: Telemetry parsers and their timestamp adapters return naive datetimes, similarly the `print_timestamp` block keeps printing naive datetimes. Eventually those interfaces could be changed to expect/return timezone-aware datetimes, but I would recommend to not do it at this point.

daniestevez:
Going through this PR and reviewing.

I have checked that the `tle_to_doppler_file.py` produces identical results with the following run
```
./examples/doppler_correction/tle_to_doppler_file.py --tle-file /tmp/tle --output-file /tmp/doppler --unix-timestamp 1759518730 --f-carrier 10e9 --lat 40 --lon=-1
```
using the following TLE
```
1 65434U 25194A   25276.07939044  .00094870  00000+0  24003-2 0  9999
2 65434  97.6012 127.9607 0003169  18.0263 342.1093 15.39396282  5869
```

python/adapters.py
```diff
@@ -49,7 +49,8 @@ def _encode(self, obj, context, path=None):
         return round(obj.timestamp())
 
     def _decode(self, obj, context, path=None):
-        return datetime.datetime.utcfromtimestamp(obj)
+        t = datetime.datetime.fromtimestamp(obj, datetime.timezone.utc)
+        return t.replace(tzinfo=None)
```
daniestevez, line 53: I think it would be fine to return here a timezone-aware timestamp with UTC timezone. Even though it's changing the API of this function, I don't think anyone is using the output of this function for anything other than printing or formatting the timestamp in human-readable format.

python/ccsds/space_packet_time_stamp_adder.py
```diff
@@ -197,7 +201,7 @@ def handle_msg(self, msg_pmt):
                 self.ascii_dec_num = 1
 
             if self.time_format == 3: # ASCII A
-                arr = self.now.isoformat()
+                arr = self.now.replace(tzinfo=None).isoformat()
```
daniestevez, line 204: Good call with `.replace(tzinfo=None)` here, since otherwise we would get the timezone formatted into the packet, which would be incorrect.

python/filereceiver/dsat.py
```diff
@@ -68,14 +68,16 @@ def _watch_file_announcements(self, chunk):
...
+            struct.unpack('<i', chunk[4:8])[0],
+            datetime.timezone.utc,
+        )
         fid = struct.unpack('<I', chunk[8:12])[0]
         # next 12 bytes are for GPS position
         length = struct.unpack('<I', chunk[21:25])[0]
 
         self.log(f'image {fid} announced. Length {length}. '
-                 f'Timestamp {timestamp}')
+                 f'Timestamp {timestamp.replace(tzinfo=None)}')
```
daniestevez, line 80: Here I think it's fine, and probably preferable, to omit `replace(tzinfo=None)`, since it is only for human-readable logging, and the presence of the timezone makes clearer that this is UTC rather than local time.

python/print_timestamp.py
```diff
@@ -34,7 +34,8 @@ def __init__(self, tstamp_format='', count_packets=False):
 
     def handle_msg(self, msg_pmt):
         if self.tstamp_format:
-            timestamp = datetime.datetime.utcnow()
+            timestamp = datetime.datetime.now(datetime.timezone.utc)
+            timestamp = timestamp.replace(tzinfo=None)
```
daniestevez, line 38: Here I think it's fine to keep the UTC timezone awareness. The timestamp is just formatted with `strftime()`, so I don't think the output will change regardless of whether there is a timezone or not, but having the timezone allows the `tstamp_format` string to use `%:z` to get the timezone printed, if the user wishes to.

python/submit.py
```diff
@@ -18,20 +19,24 @@
...
+        t_int = datetime.datetime.fromisoformat(s_int)
+        t_frac = float(f'0.{s_frac}')
+        # Truncate t_frac to integer microseconds to match
+        # behaviour of fromisoformat() in newer Python version
+        t = t_int + datetime.timedelta(
+            microseconds=int(t_frac * 1000000))
+    else:
+        t = datetime.datetime.fromisoformat(s)
+    t = t.replace(tzinfo=datetime.timezone.utc)
+    return t
```
daniestevez, line 39: I have done more testing and spotted an obvious bug in the Python < 3.11 implementation: the timestamp may not have a fractional part!

This is a possible fix:
```suggestion
def parse_timestamp(s):
    t = None
    if sys.version_info < (3, 11):
        # in Python 3.10, fromisoformat() does not accept a trailing Z, and it
        # can only parse millisecond resolution. If there is fractional part, we
        # remove any trailing Z, and parse the fractional seconds separately
        s = s.rstrip('Z')
        if '.' in s:
            s_split = s.split('.')
            s_int, s_frac = s.split('.')
            t_int = datetime.datetime.fromisoformat(s_int)
            t_frac = float(f'0.{s_frac}')
            # Truncate t_frac to integer microseconds to match behaviour of
            # fromisoformat() in newer Python version
            t = t_int + datetime.timedelta(
                microseconds=int(t_frac * 1000000))
    if t is None:
        t = datetime.datetime.fromisoformat(s)
    t = t.replace(tzinfo=datetime.timezone.utc)
    return t
```

python/pwsat2_submitter.py
```diff
@@ -62,8 +62,9 @@ def __init__(self, credentials_file, initialTimestamp):
         dtformat = '%Y-%m-%d %H:%M:%S'
         self.initialTimestamp = (
             datetime.datetime.strptime(initialTimestamp, dtformat)
+            .replace(tzinfo=datetime.timezone.utc)
```
daniestevez, line 65: I have just added the following test case to my script:
```python
    t3 = parse_timestamp('2025-10-02 20:50:03')
    assert str(t3) == '2025-10-02 20:50:03+00:00'
```
It works fine for all Python >= 3.8. Note that the timestamp I have used matches the format expected by `dtformat`. So we could use `parse_timestamp()` rather than `strptime()`  and be more general in the timestamp formats that we accept as `initialTimestamp`.

python/telemetry/cirbe_70cm.py
```diff
@@ -42,8 +43,16 @@ def _encode(self, obj, context, path=None):
...
+        t = datetime.datetime.fromtimestamp(
+            obj.time_stamp_seconds,
+            tz=datetime.timezone.utc
+        )
+        t = t.replace(tzinfo=None)
+        offset = (
+            datetime.datetime(2000, 1, 1, 12)
+            - datetime.datetime(1970, 1, 1)
+        )
+        return (t + offset)
```
daniestevez, line 55: I'm looking at what the original code did and it is a badly written (but correct) implementation of getting a `datetime` that is `obj.time_stamp_seconds` after the J2000 epoch (which is 2000-01-01T12:00:00Z). Therefore, the following implementation seems much more clear:
```python
j2000_epoch = datetime.datetime(2000, 1, 1, 12, tzinfo=datetime.datetime.utc)
return j2000_epoch + datetime.timedelta(seconds=obj.time_stamp_seconds)
```

I would probably return a timezone-aware datetime here.

python/telemetry/cute_70cm.py
```diff
@@ -32,25 +33,6 @@
     )
 
 
-class TimeAdapter(Adapter):
```
daniestevez, line 35: Well spotted code duplication. It's probably my fault during PR reviews for PRs from this satellite team.

daniestevez (commented):
Thanks so much for going through all the codebase making these changes and for the other clean up!

I think that it would be fine to return timezone aware objects in all places, and I have left comments where I think that we can omit a `.replace(tzinfo=None)`. But I leave it up to you if you want to do it in this PR. Otherwise I can do it on a follow-up PR.

daniestevez:
> This is fixed by using t.astimezone(datetime.timezone.utc):

Good catch, but using `astimezone()` isn't right either, because it converts naïve datetimes by assuming that they are in local time rather than UTC.

For instance,
```
>>> str(datetime.datetime.fromisoformat('2025-01-01 00:00').astimezone(datetime.timezone.utc))
'2024-12-31 23:00:00+00:00'
```
which is wrong, because in most contexts where we would call `parse_timestamp()`, this `2025-01-01 00:00` without an explicit timezone would mean UTC rather than my local timezone (CET).

Probably we need something like
```python
if t.tzinfo is None:
   # assume UTC if no timezone was specified in the string
    t = t.replace(tzinfo=datetime.timezone.utc)
```
and then we might either do `t = t.astimezone(datetime.timezone.utc)` as you suggested if we really want this function to always return aware datetimes in UTC (which might simplify the case analysis of the rest of the code), or omit converting the timezone if we want this function to allow returning aware datetimes in other timezones if the user specified that (which probably would work, but makes the analysis more complex).

Tomorrow I'll think more carefully about how to solve this, covering also the Python < 3.11 case if possible.

kerel-fs:
> [..] using `astimezone()` isn't right either, because it converts naïve datetimes by assuming that they are in local time rather than UTC.

This is an unexpected default that I missed, introduced in Python 3.6.

I was hoping we could avoid having unit tests for `parse_timestamp`, but I slowly begin to think having your test script as an actual unit test in gr-satellites could be helpful in the future. What do you think?

daniestevez:
I agree that it would be good to have a unit test for `parse_timestamp()`. I was thinking that timestamp handling across Python versions is quite confusing, so ideally we would have unit tests for all the blocks that use timestamps and run these tests in all Python versions >= 3.8. However this is unrealistic, because we would need to build GNU Radio for each of these Python versions. But putting the test script I've been using as a unit test for `parse_timestamp()` is easy to do and seems worth the effort.

I have done some more work on this, testing other weird corner cases such as timezones with a fractional second part (which are properly supported in Python >= 3.11). I have managed to write an implementation that also works fine for Python < 3.11. I'm not thrilled by how this function has grown to consider all the possible cases, but this is the better option we have for now. My script is now like this:

```python
#!/usr/bin/env python3

import datetime
import sys


def parse_timestamp(s):
    t = None
    t_tz_frac = None
    if sys.version_info < (3, 11):
        # In Python 3.10, fromisoformat() does not accept a trailing Z, and it
        # can only parse millisecond resolution. If there is fractional part, we
        # remove any trailing Z, and parse the fractional seconds
        # separately. Python 3.10 does not support fractional seconds in the
        # timezone, so this is also handled separately.
        s = s.rstrip('Z')
        if '+' in s:
            # s has a timezone
            s, s_tz = s.split('+')
            if '.' in s_tz:
                # The timezone has a fractional seconds part. The fractional
                # part is incorporated back near the end of this function.
                s_tz, s_tz_frac = s_tz.split('.')
                t_tz_frac = float(f'0.{s_tz_frac}')
        else:
            s_tz = None
        if '.' in s:
            # s has a fractional seconds part
            s_int, s_frac = s.split('.')
            if s_tz is not None:
                # Put back the timezone in s_int
                s_int = f'{s_int}+{s_tz}'
            t_int = datetime.datetime.fromisoformat(s_int)
            t_frac = float(f'0.{s_frac}')
            # Truncate t_frac to integer microseconds to match behaviour of
            # fromisoformat() in newer Python version
            t = t_int + datetime.timedelta(
                microseconds=int(t_frac * 1000000))
        elif s_tz is not None:
            # Put back the timezone in s
            s = f'{s}+{s_tz}'
    if t is None:
        t = datetime.datetime.fromisoformat(s)
    if t_tz_frac is not None:
        t -= datetime.timedelta(microseconds=int(t_tz_frac * 1000000))
    if t.tzinfo is None:
        # The default timezone for this function is UTC
        t = t.replace(tzinfo=datetime.timezone.utc)
    else:
        # Make sure that we are returning a datetime object for the UTC
        # timezone, to avoid surprises by mixing timezones in downstream code
        t = t.astimezone(datetime.timezone.utc)
    return t


def main():
    # parse timestamps in various formats
    t0 = parse_timestamp('2025-10-02T17:49:22.123456789Z')
    assert str(t0) == '2025-10-02 17:49:22.123456+00:00'

    t1 = parse_timestamp('2025-10-02T20:47:24.587123442')
    assert str(t1) == '2025-10-02 20:47:24.587123+00:00'

    t2 = parse_timestamp('2025-10-02T20:50:03.157291322Z')
    assert str(t2) == '2025-10-02 20:50:03.157291+00:00'

    t3 = parse_timestamp('2025-10-02 20:50:03')
    assert str(t3) == '2025-10-02 20:50:03+00:00'

    t4 = parse_timestamp('2025-10-02T17:49:22+02:00')
    assert str(t4) == '2025-10-02 15:49:22+00:00'

    t5 = parse_timestamp('2025-10-02T17:49:22+02:30')
    assert str(t5) == '2025-10-02 15:19:22+00:00'

    t6 = parse_timestamp('2025-10-02T17:49:22+02:00:00.01')
    assert str(t6) == '2025-10-02 15:49:21.990000+00:00'

    # get current time
    t_now = datetime.datetime.now(tz=datetime.timezone.utc)
    print(f'now = {t_now}')

    # get elapsed time
    elapsed = (t1 - t0).total_seconds()
    assert elapsed == 3 * 3600 - 2 * 60 + 2 + 0.587123 - 0.123456

    # propagate time (as in submit.py)
    t_prop = t2 - t1 + t0
    assert str(t_prop) == '2025-10-02 17:52:00.693624+00:00'

    # format as needed in submit.py
    t0_fmt = t0.replace(tzinfo=None).isoformat()[:-3] + 'Z'
    assert t0_fmt == '2025-10-02T17:49:22.123Z'

    # compute KISS timestamp
    t0_kiss = round(t0.timestamp() * 1e3)
    assert t0_kiss == 1759427362123


if __name__ == '__main__':
    main()
```

python/adapters.py
```diff
@@ -46,10 +46,11 @@ def _decode(self, obj, context, path=None):
 
 class UNIXTimestampAdapter(Adapter):
     def _encode(self, obj, context, path=None):
-        return round(obj.timestamp())
+        return round(obj.astimezone(datetime.timezone.utc).timestamp())
```
daniestevez: I like the `astimezone()` here, as it will make a naïve `datetime` be interpreted as UTC rather than as local time. For aware `datetime`s, I think the `astimezone()` here doesn't change the output.

daniestevez (approved):
This looks great now. Thanks for all the work! Approved and merging.

Issue #770, opened 2025-10-01
ERMINAZ: Add support for new framing


kerel-fs:
The ERMINAZ team modified their downlink protocol to better follow the CCSDS standards.

**FSK demodulation & Deframing**

Changes (non-exhaustive list, I might be missing items):
- Fixed preamble
- Fixed syncword
- Flipped FEC & Scrambling
- Changed framing to "CCSDS TM Frames (128 octets including FEC) with encapsulated Space Packets of variable size (0 to 65535 octets)", see [ERMINAZ Data Formats](https://gitlab.com/amsat-dl/erminaz/erminaz-1-org/-/wikis/Erminaz-Commands)

A modified SatYAML (see https://github.com/daniestevez/gr-satellites/pull/771) as well as an IQ recording was shared (285MB, `recording_435775000Hz_20-43-13_20-09-2025.sigmf-data/meta` / todo: add URLs to IQ files once published on a suitable webserver).

I was able to decode up to 684 Frames from this recording, with 47 / 5 / 2 / 2 **lost frames** identified by checking the `master_channel_frame_count` in the TM Transfer Frame Primary Header, using a relative clock recovery bandwidth of 0.06 (default) / 0.1 / 0.15 / 0.2. Given the high SNR of this recording, I would expect no frame loss at all, so something still seems slightly off.

To investigate further, I created SigMF annotations based on the timestamps in the output KISS file. This allows us to identify with `inspectrum` when frames are not decoded. One of those missing frames begins at `t=44.73 s`:

<img width="1920" height="1048" alt="Image" src="https://github.com/user-attachments/assets/addac08c-c634-48bf-a291-7313ca543fa6" />

The previous and following frames are decoded, only this single frame is missing. I dumped & plotted some internal signals for this segment:
<details><summary>Dumped internal signals</summary>
<p>
<img width="1920" height="971" alt="Image" src="https://github.com/user-attachments/assets/8589fb04-6d43-4d1c-9b2e-e78ce91f9dd4" />
<img width="1920" height="971" alt="Image" src="https://github.com/user-attachments/assets/dd0cc175-6f1e-4974-a21e-942bb7825b1a" />
<img width="1920" height="971" alt="Image" src="https://github.com/user-attachments/assets/655d98f9-a64e-4593-8049-441aebdfd9f1" />
<img width="1920" height="971" alt="Image" src="https://github.com/user-attachments/assets/8c865408-1ac6-41de-a2f3-923646e50fb7" />
[...]

PR #771, opened 2025-10-01
Update ERMINAZ-1U.yml

kerel-fs:
Framing was changed to follow CCSDS standards more closely.

See also #770.

### DIFF
--- a/python/satyaml/ERMINAZ-1U.yml
+++ b/python/satyaml/ERMINAZ-1U.yml
@@ -4,11 +4,24 @@ data:
   &tlm Telemetry:
     telemetry: erminaz
 transmitters:
+  # Old framing:
   9k6 FSK downlink:
     frequency: 435.775e+6
     modulation: FSK
     baudrate: 9600
     deviation: 3000
     framing: QUBIK
+    data:
+    - *tlm
+  # New framing:
+  9k6 FSK CCSDS TMTC SPP downlink:
+    frequency: 435.775e+6
+    modulation: FSK
+    baudrate: 9600
+    deviation: 4800
+    framing: CCSDS Reed-Solomon
+    frame size: 128
+    RS basis: conventional
+    scrambler: CCSDS
     data:
     - *tlm


python/satyaml/ERMINAZ-1U.yml
```diff
@@ -12,3 +13,17 @@ transmitters:
     framing: QUBIK
     data:
     - *tlm
+  # New framing:
+  9k6 FSK CCSDS TMTC SPP downlink:
+    frequency: 435.775e+6
+    modulation: FSK
+    baudrate: 9600
+    deviation: 4800
```
daniestevez, line 21: The old 9k6 framing has 3000 Hz deviation. Has the deviation been changed too?

python/satyaml/ERMINAZ-1U.yml
```diff
@@ -12,3 +13,17 @@ transmitters:
...
+    framing: CCSDS Reed-Solomon
+    frame size: 128
+    RS basis: conventional
+    RS interleave depth: 1
```
daniestevez, line 25: It is not necessary to specify the `RS interleave depth`, since the default is one.

python/satyaml/ERMINAZ-1U.yml
```diff
@@ -12,3 +13,17 @@ transmitters:
...
+    scrambler: CCSDS
+    syncword threshold: 4
```
daniestevez, line 27: I am not sure that this field has any effect. `gr_satellites_flowgraph.py` does not appear to read it. The syncword threshold is handled by a command line argument.

python/satyaml/ERMINAZ-1U.yml
kerel-fs: @DL4PD Has the deviation changed for ERMINAZ?

python/satyaml/ERMINAZ-1U.yml
kerel-fs: Fixed (removed).

Moreover, the correct keyword to specify a non-default R-S interleaving depth would have been `RS interleaving`.

daniestevez:
The changes look good now pending @DL4PD to confirm if the deviation is 4800 Hz or 3000 Hz.

python/satyaml/ERMINAZ-1U.yml
kerel-fs: Confirmed, the deviation was changed to 4800 Hz to match the configuration in the firmware.

ERMINAZ uses the AX5043 radio, and in [src/ax5043.c#L559-582](https://gitlab.com/librespacefoundation/ax5043-driver/-/blob/f3589fd9dd5ce2ac19aa56053fd8064807c20613/src/ax5043.c#L559-582) the deviation frequency  $f_{dev}$ is set to $f_{dev} = h / 2 * f_{baud}$, based on the modulation index $h$ and baudrate $f_{baud}$. This method is called by `int ax5043_conf_tx(...)` at [Core/Src/radio.c#L308-350](https://gitlab.com/amsat-dl/erminaz/erminaz-comms-sw/-/blob/4ebeeb4467919dff1373a4babab43a9cef7c807b/Core/Src/radio.c#L308-350) in the firmware (where $h = 1, Ff_{baud} = 9600$). Assuming this method is implemented correctly, the deviation thus should be 4800 Hz for a 9600 baud FSK transmission.

I don't know why it was set to 3000 Hz originally in the SatYAML, it might have been a guess?

daniestevez: This looks good. Let's set the deviation to 4800 Hz, then.

The 3000 Hz value comes from https://destevez.net/2024/06/decoding-erminaz/

> The modulation is simple enough: 9600 baud GFSK with a deviation of 3000 Hz.

I stated this without a reference, so I don't know where I got this value from. It is quite likely that I measured this in some of the pre-flight recordings that I was given.

PR #773, opened 2025-10-06
ERMINAZ: Fix SSDV reception

kerel-fs:
The ERMINAZ team changed their downlink protocol,  in this patch I'm addressing the remaining changes to fix SSDV reception.

SSDV packets are 256 bytes now, encapsulated into CCSDS Space Packets and transfered in 128 bytes fixed-length Transfer Frames.

I had already implemented a new packet extraction method before I found the existing [`telemetry_packet_reconstruction` block](https://github.com/daniestevez/gr-satellites/blob/main/python/ccsds/telemetry_packet_reconstruction.py). Thus, this pull request is considered a draft, and I'll have to address the code duplication before it is ready for review.

### DIFF
--- a/examples/ccsds/TM_Packet_with_Space_Packet.grc
+++ b/examples/ccsds/TM_Packet_with_Space_Packet.grc
@@ -47,7 +47,7 @@ blocks:
     bus_sink: false
     bus_source: false
     bus_structure: null
-    coordinate: [80, 132]
+    coordinate: [64, 128.0]
     rotation: 0
     state: enabled
 - name: pdu_random_pdu_0
@@ -66,7 +66,7 @@ blocks:
     bus_sink: false
     bus_source: false
     bus_structure: null
-    coordinate: [472, 36.0]
+    coordinate: [352, 112.0]
     rotation: 0
     state: true
 - name: satellites_ccsds_telemetry_parser_0
@@ -81,7 +81,7 @@ blocks:
     bus_sink: false
     bus_source: false
     bus_structure: null
-    coordinate: [632, 424]
+    coordinate: [200, 360.0]
     rotation: 0
     state: enabled
 - name: satellites_space_packet_parser_0
@@ -98,7 +98,7 @@ blocks:
     bus_sink: false
     bus_source: false
     bus_structure: null
-    coordinate: [880, 412]
+    coordinate: [680, 352.0]
     rotation: 0
     state: enabled
 - name: satellites_space_packet_primaryheader_adder_0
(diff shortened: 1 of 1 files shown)


kerel-fs:
From the recording shared in #770, 684 transfer frames can be decoded: [erminaz_2026-09-20.kss.zip](https://github.com/user-attachments/files/22726904/erminaz_2026-09-20.kss.zip)

```
❯ kiss_info /tmp/erminaz_2026-09-20.kss
Input File: /tmp/erminaz_2026-09-20.kss
Total:       684 frames

Frame sizes:
  128 bytes - 684 frames

First timestamp: 2025-09-20 20:43:14.777000+00:00
Last timestamp:  2025-09-20 20:45:52.470000+00:00
```

There are 16 SSDV packets for a single image (`image_id=0x55`) in this recording:
```
❯ gr_satellites_ssdv --satellite ERMINAZ /tmp/erminaz_2026-09-20.kss /tmp/ssdv
Calling SSDV decoder for image 0x55
Callsign: DP0SAT
Image ID: FF
Resolution: 128x112
MCU blocks: 224
Sampling factor: 1x1
Quality level: 4
Read 16 packets
```

kerel-fs:
Issue encountered with the original `telemetry_packet_reconstruction` block:
- No detection of frame loss, thus recovery of the packet reconstruction function after a lost frame is delayed.
- No idle frame detection
- TM Frames created with the `TM_Packet_with_Space_Packet.grc` flowgraph seem to have an unexpected size (to be investigated further)

I'd like to add frame loss detection as optional feature:
> reset_on_frame_loss (bool): If enabled, check virtual channel frame counter to detect lost frames.

tbd: Check why "TM_Packet_with_Space_Packet.grc" fails with the new block implementation (even if frame loss detection is disabled).

daniestevez:
The original `telemetry_packet_reconstruction` block was written by a Google Summer of Code student, alongside other blocks to process CCSDS. It hasn't been used much or maintained much. So I wouldn't be surprised that it has shortcomings, and it seems fine to me to throw away (ignore) this code and write something better.

Ideally we would have a gr-satellites transport (just like the [KISS transport](https://github.com/daniestevez/gr-satellites/blob/main/python/components/transports/kiss_transport.py)) that defragments CCSDS Space Packets.

I have [the following code](https://github.com/daniestevez/jupyter_notebooks/blob/master/Tianwen/ccsds.py#L67) to defragment space packets. It supports both TM Space Data Link frames and AOS frames, and it has been tested with many deep space missions and in different cases of frame loss and formatting subtleties (such as part of the packet zone being filled with padding rather than idle Space Packets), so I can vouch that this code is very robust. This code could directly go into a CCSDS Space Packet transport. The `construct` structures that are needed for parsing would go with the rest of the telemetry parsers in the `python/telemetry` directory (we probably already have some of these in gr-satellites).

PR #831, opened 2026-05-27
Add length checks in packet cropping

daniestevez:
When cropping a packet, the call to pmt.init_u8vector(len, data) can result in copying out-of-bounds data if len(data) < len. This adds lengths checks to the cc11x_packet_crop, sx12xx_packet_crop and spino_crop. Additionally, the maximum packet length (used by the sync_to_pdu_packed block) of deframers that use a cc11x_packet_crop or sx12xx_packet_crop block with 2 bytes of CRC has been increased to 258 bytes, so that it's impossible to run into this case.

### DIFF
--- a/python/sx12xx_packet_crop.py
+++ b/python/sx12xx_packet_crop.py
@@ -40,6 +40,11 @@ class sx12xx_packet_crop(gr.basic_block):
         packet = bytes(pmt.u8vector_elements(msg))
 
         packet_length = packet[0] + 1 + self.crc_len
+        if packet_length > len(packet):
+            print(
+                f'packet length {packet_length} greater than '
+                f'PDU size {len(packet)}; dropping)')
+            return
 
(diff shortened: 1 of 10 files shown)


python/sx12xx_packet_crop.py
```diff
@@ -40,6 +40,11 @@ def handle_msg(self, msg_pmt):
         packet = bytes(pmt.u8vector_elements(msg))
 
         packet_length = packet[0] + 1 + self.crc_len
+        if packet_length > len(packet):
+            print(
+                f'packet length {packet_length} greater than '
+                f'PDU size {len(packet)}; dropping)')
```
Marnick39, line 46: Cosmetic: there's a stray `)` in the sx12xx drop-message string

daniestevez: Good catch! Addressed in #834.

PR #840, opened 2026-07-03
Add MIST support

william-stacken:
See https://github.com/daniestevez/gr-satellites/discussions/837

### DIFF
--- a/python/satyaml/MIST.yml
+++ b/python/satyaml/MIST.yml
@@ -1,0 +1,34 @@
+name: MIST
+norad: ~
+data:
+  &tlm Telemetry:
+    telemetry: ax25
+transmitters:
+  1k2 BPSK downlink:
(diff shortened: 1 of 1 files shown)


python/satyaml/MIST.yml
```diff
@@ -0,0 +1,34 @@
+name: MIST
+norad: ~
```
daniestevez, line 2: This needs to contain a number https://github.com/daniestevez/gr-satellites/actions/runs/28659018391/job/84994933247?pr=840#step:4:614

You should request a temporary NORAD ID from SatNOGS DB.

william-stacken:
@daniestevez Sorry for the wait! We have now obtained a temporary NORAD ID from SatNOGS DB.

daniestevez (approved):
Looks good. Merging.

PR #841, opened 2026-07-04
Commit has the following changes

PrajinIr:
1. We have implemented the decoder for cubesat named KOYO.
2. Added corresponding yml file.
3.  Added test wav file and tested.

### DIFF
--- a/python/satyaml/KOYO.yml
+++ b/python/satyaml/KOYO.yml
@@ -1,5 +1,5 @@
 name: KOYO
-norad: 799602095
+norad: 98273
 data:
   &tlm Telemetry:
     telemetry: koyo
@@ -8,7 +8,7 @@ transmitters:
     frequency: 435.400e+6
     modulation: FSK
     baudrate: 9600
-    deviation: 4800
+    deviation: 2400
     framing: AX.25 G3RUH
     data:
     - *tlm
--- a/python/satyaml/KOYO.yml
+++ b/python/satyaml/KOYO.yml
@@ -1,0 +1,14 @@
+name: KOYO
+norad: 799602095
+data:
+  &tlm Telemetry:
+    telemetry: koyo
+transmitters:
+  9k6 FSK downlink:
+    frequency: 435.400e+6
+    modulation: FSK
+    baudrate: 9600
+    deviation: 4800
+    framing: AX.25 G3RUH
+    data:
+    - *tlm
--- a/python/telemetry/CMakeLists.txt
+++ b/python/telemetry/CMakeLists.txt
@@ -82,6 +82,7 @@ GR_PYTHON_INSTALL(
     tubix20.py
     upmsat_2.py
     vzlusat_2.py
+    koyo.py
     DESTINATION ${GR_PYTHON_DIR}/satellites/telemetry
 )
 
--- a/python/telemetry/CMakeLists.txt
+++ b/python/telemetry/CMakeLists.txt
@@ -64,6 +64,7 @@ GR_PYTHON_INSTALL(
     gomx_1.py
     gomx_3.py
     inspiresat_1.py
+    koyo.py
     kr01.py
     lume.py
     mirsat1.py
@@ -82,7 +83,6 @@ GR_PYTHON_INSTALL(
     tubix20.py
     upmsat_2.py
     vzlusat_2.py
-    koyo.py
     DESTINATION ${GR_PYTHON_DIR}/satellites/telemetry
 )
 
--- a/python/telemetry/__init__.py
+++ b/python/telemetry/__init__.py
@@ -65,5 +65,6 @@ from .trisat import trisat
 from .tubix20 import tubix20
 from .upmsat_2 import upmsat_2
 from .vzlusat_2 import vzlusat_2
+from .koyo import koyo
 
 construct.setGlobalPrintFullStrings(True)
--- a/python/telemetry/__init__.py
+++ b/python/telemetry/__init__.py
@@ -44,6 +44,7 @@ from .funcube import funcube
 from .gomx_1 import gomx_1
 from .gomx_3 import gomx_3
 from .inspiresat_1 import inspiresat_1
+from .koyo import koyo
 from .kr01 import kr01
 from .lume import lume
 from .mirsat1 import mirsat1
@@ -65,6 +66,5 @@ from .trisat import trisat
 from .tubix20 import tubix20
 from .upmsat_2 import upmsat_2
 from .vzlusat_2 import vzlusat_2
-from .koyo import koyo
 
 construct.setGlobalPrintFullStrings(True)
--- a/python/telemetry/koyo.py
+++ b/python/telemetry/koyo.py
@@ -1,33 +1,33 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
- 
+
 # Copyright 2026 Hex20 Labs India Pvt Ltd
 #
 # This file is part of gr-satellites
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 #
+
 from .ax25 import Header
 from ..adapters import PolynomialAdapter
 from ..adapters import LinearAdapter
+from ..adapters import UNIXTimestampAdapter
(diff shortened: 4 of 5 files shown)


python/satyaml/KOYO.yml
```diff
@@ -0,0 +1,14 @@
+name: KOYO
+norad: 799602095
```
daniestevez: I see that this satellite has been proposed as an addition to SatNOGS DB: https://db.satnogs.org/satellite/YZNT-7399-1272-5956-2962 and that it has the same NORAD ID there. However, this NORAD ID may change, since it does not follow the usual 5-digit format. We should hold off merge until this has been added to SatNOGS DB and the temporary NORAD ID is confirmed.

python/satyaml/KOYO.yml
```diff
@@ -0,0 +1,14 @@
...
+data:
+  &tlm Telemetry:
+    telemetry: koyo
+transmitters:
+  9k6 FSK downlink:
+    frequency: 435.400e+6
+    modulation: FSK
+    baudrate: 9600
+    deviation: 4800
```
daniestevez: Can you confirm that the deviation is 4800 Hz? GFSK would be 2400 Hz deviation. Do you have an IQ recording where this can be measured.

python/telemetry/__init__.py
```diff
@@ -65,5 +65,6 @@
 from .tubix20 import tubix20
 from .upmsat_2 import upmsat_2
 from .vzlusat_2 import vzlusat_2
+from .koyo import koyo
```
daniestevez: This should be in alphabetic order.

python/telemetry/CMakeLists.txt
```diff
@@ -82,6 +82,7 @@ GR_PYTHON_INSTALL(
     tubix20.py
     upmsat_2.py
     vzlusat_2.py
+    koyo.py
```
daniestevez: This should be in alphabetic order.

python/telemetry/koyo.py
```diff
@@ -0,0 +1,146 @@
...
+    87.1751343, -0.0786252941, 0.0000272861362, -0.00000000402689014]
+subsystem_temp_poly = [91.394, -0.08949, 3.55e-05, -6.26e-09, 1.89e-13]
+
+
+SecondaryHeader = Struct(
+    'sh_coarse' / Int32ul,
+    'sh_fine' / Int16ul,
+)
+ 
+koyo_beacon = Struct(
```
daniestevez, line 29: Is there public documentation for the telemetry which is implemented here? If there is, we should add a link to it in a comment here. Otherwise this could can serve as public documentation.

python/telemetry/koyo.py
```diff
@@ -0,0 +1,146 @@
...
+ 
+solar_panel_temp_poly = [91.394, -0.08949, 3.55e-05, -6.26e-09, 1.89e-13]
+batt_thermistor_temp_poly = [
+    87.1751343, -0.0786252941, 0.0000272861362, -0.00000000402689014]
+subsystem_temp_poly = [91.394, -0.08949, 3.55e-05, -6.26e-09, 1.89e-13]
+
+
+SecondaryHeader = Struct(
+    'sh_coarse' / Int32ul,
+    'sh_fine' / Int16ul,
```
daniestevez: What is `sh_`? Is it "secondary header"? Why does it have coarse and fine? Is it supposed to be a timestamp?

python/telemetry/koyo.py
```diff
@@ -0,0 +1,146 @@
...
+    'ax25_header' / Header,
+    'primary_header' / ccsds_space_packet.PrimaryHeader,
+    'secondary_header' / If(
+        lambda c: c.primary_header.secondary_header_flag,
+        SecondaryHeader
+    ),
+    'packet' / Switch(
+        lambda c: c.primary_header.AP_ID,
+        {
+            0x01: koyo_beacon,
```
daniestevez, line 142: Are there any other APIDs used by the satellite?

python/telemetry/koyo.py
```diff
@@ -0,0 +1,146 @@
...
+subsystem_temp_poly = [91.394, -0.08949, 3.55e-05, -6.26e-09, 1.89e-13]
+
+
+SecondaryHeader = Struct(
+    'sh_coarse' / Int32ul,
+    'sh_fine' / Int16ul,
+)
+ 
+koyo_beacon = Struct(
+    'obctime' / Int32ul,
```
daniestevez: This looks like a UNIX timestamp in the WAV recording. We have a [UNIXTimestampAdapter](https://github.com/daniestevez/gr-satellites/blob/main/python/adapters.py#L47) that could be used here.

daniestevez (changes requested):
Thanks for submitting this PR! In general it looks good, but I have some minor comments. Also, make sure that pycodestyle is passing.

python/satyaml/KOYO.yml
PrajinIr: image

799602095 is the NORAD-CAT ID assigned for us in SPACE-TRACK.ORG. So we used this as the NORAD ID. Please confirm whether it is acceptable.

python/telemetry/__init__.py
PrajinIr: Updated as per the comment.

python/telemetry/CMakeLists.txt
PrajinIr: Updated as per the comment.

python/telemetry/koyo.py
PrajinIr: Secondary header represents millisecond value and day value. Variable name changed accordingly.

python/telemetry/koyo.py
PrajinIr: By default we have only single beacon packet with APID 0x01.

python/telemetry/koyo.py
PrajinIr: Updated as per comment.

python/satyaml/KOYO.yml
daniestevez: Sounds good. There is a very slight chance that it causes issues with something downstream that expects 5-character NORAD IDs, but I'm not going to contradict Space-Track.

python/telemetry/koyo.py
PrajinIr: We are not keeping a public documentation for the telemetry right now.

python/satyaml/KOYO.yml
PrajinIr: Deviation will be 2400 Hz if the modulation index is 0.5, but we considered the modulation index value as 1 in-order to get better noise immunity. In that case the deviation value is set as 4800.

daniestevez:
There is something strange going on with the satellite-recordings submodule in this PR.

Screenshot from 2026-07-05 22-18-59

daniestevez:
https://github.com/daniestevez/satellite-recordings/pull/8 is merged, you should update the satellite-recordings submodule reference in this PR to point to the latest master.

python/satyaml/KOYO.yml
PrajinIr: NORAD ID updated since temporary NORAD ID is confirmed.

PrajinIr:
> [daniestevez/satellite-recordings#8](https://github.com/daniestevez/satellite-recordings/pull/8) is merged, you should update the satellite-recordings submodule reference in this PR to point to the latest master.

Updated as per comment

PrajinIr:
> There is something strange going on with the satellite-recordings submodule in this PR.
> Screenshot from 2026-07-05 22-18-59

Could you please confirm this issue is resolved.

daniestevez (approved):
Looks good. Squashing and merging.

python/satyaml/KOYO.yml
daniestevez: @PrajinIr I've noticed you've changed this to 2400. What is the reason?

PrajinIr: We discussed with the COMM team and confirmed that modulation index for KOYO had been set as 0.5.
Sorry for the misunderstanding happened.

PR #846, opened 2026-08-12
Pass real point-in-file through during replay from wav file for accurate timestamps

### LINKED ISSUE #92: Add timestamp handling for telemetry submission in recordings
Currently the timestamp for telemetry submission is taken from the system time, so recordings don't have the correct timestamp. By adding and tracking timestamps appropriately in the data source, we could have good timestamp tagging arriving into the telemetry submit block. However, this doesn't seem trivial.

parker-research:
Fixes #92 (the oldest currently-open issue). 

This has been tested a fair bit, though focusing on decoding from wav files. I haven't tested it with live decoding or anything, though I don't suspect there'll be any regressions for those.

The only possible "regression" is in the case of using `--start_time` with a live/hardware source. Previously, that configuration spits out realtime timestamps. Now, it'll produce timestamps based on the sample rate and sample count, which may have different behaviour than previously. I suspect there aren't a lot of people using `--start_time` with a live/hardware source, so it's probably not a huge deal in reality. It is my opinion that this is better, and shouldn't be considered a regression.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -747,15 +747,25 @@ Getting correct timestamps with recordings
 One of the difficulties with working with recordings is obtaining correct
 timestamps for each of the decoded packets. These timestamps are included in
 KISS files and telemetry submissions to some servers, such as SatNOGS DB. To
-produced correct timestamps ``gr_satellites`` will play back the recording at 1x
-speed and count the clock time elapsed since the beginning of the execution, it
-will then add that time to a timestamp specified by the user, which should
-correspond to the start of the recording.
+produce correct timestamps, always use the ``--start_time`` parameter followed
+by the timestamp in ISO 8601 format (``YYYY-MM-DDTHH:MM:SS``) to indicate the
+start time of the recording.
 
-To use this functionality it is necessary to use the ``--throttle`` parameter to
-limit playback speed to 1x and use the ``--start_time`` parameter followed by the
-timestamp in ISO 8601 format (``YYYY-MM-DDTHH:MM:SS``) to indicate the start time
-of the recording.
+Many deframers (those based on ``sync_to_pdu``, such as most fixed-length
+framings, and the AX.25 deframer) attach the offset of each frame within the
+recording to the decoded packet. When this information is available,
+``gr_satellites`` computes each packet's timestamp directly from that offset
+and the transmitter's baudrate, so the recording can be decoded as fast as
+possible while still producing correct timestamps; ``--throttle`` is not
+needed in this case.
+
+For deframers that do not yet attach this information, ``gr_satellites``
+falls back to playing back the recording at 1x speed and counting the clock
+time elapsed since the beginning of the execution, adding that elapsed time
+to the timestamp given by ``--start_time``. To use this fallback it is
+necessary to use the ``--throttle`` parameter to limit playback speed to 1x
+(this is harmless to leave enabled even when the faster, offset-based
+timestamping applies).
 
 Treating unknown args as warning
 """"""""""""""""""""""""""""""""
--- a/lib/ax100_decode_impl.cc
+++ b/lib/ax100_decode_impl.cc
@@ -90,7 +90,7 @@ void ax100_decode_impl::msg_handler(pmt::pmt_t pmt_msg)
         // Send by GNUradio message
         message_port_pub(
             pmt::mp("out"),
-            pmt::cons(pmt::PMT_NIL, pmt::init_u8vector(frame_len, &d_data[1])));
+            pmt::cons(pmt::car(pmt_msg), pmt::init_u8vector(frame_len, &d_data[1])));
     } else if (d_verbose) {
         std::printf("RS decode failed.\n");
     }
--- a/lib/decode_ra_code_impl.cc
+++ b/lib/decode_ra_code_impl.cc
@@ -107,7 +107,7 @@ void decode_ra_code_impl::msg_handler(pmt::pmt_t pmt_msg)
     if ((float)errors / (ra_code_length * RA_BITCOUNT) < d_error_threshold) {
         message_port_pub(
             pmt::mp("out"),
-            pmt::cons(pmt::PMT_NIL, pmt::init_u8vector(d_size, d_ra_out.data())));
+            pmt::cons(pmt::car(pmt_msg), pmt::init_u8vector(d_size, d_ra_out.data())));
     }
 }
 
--- a/lib/distributed_syncframe_soft_impl.cc
+++ b/lib/distributed_syncframe_soft_impl.cc
@@ -66,13 +66,11 @@ int distributed_syncframe_soft_impl::work(int noutput_items,
         if (match >= d_syncword.size() - d_threshold) {
             // sync found
             pmt::pmt_t meta = pmt::make_dict();
-            meta = pmt::dict_add(meta,
-                                 pmt::mp("sample_offset"),
-                                 pmt::from_uint64(nitems_read(0) + i));
+            meta = pmt::dict_add(
+                meta, pmt::mp("sample_offset"), pmt::from_uint64(nitems_read(0) + i));
             message_port_pub(
                 pmt::mp("out"),
-                pmt::cons(meta,
-                          pmt::init_f32vector(d_syncword.size() * d_step, in + i)));
+                pmt::cons(meta, pmt::init_f32vector(d_syncword.size() * d_step, in + i)));
         }
     }
 
--- a/lib/distributed_syncframe_soft_impl.cc
+++ b/lib/distributed_syncframe_soft_impl.cc
@@ -65,9 +65,13 @@ int distributed_syncframe_soft_impl::work(int noutput_items,
         }
         if (match >= d_syncword.size() - d_threshold) {
             // sync found
+            pmt::pmt_t meta = pmt::make_dict();
+            meta = pmt::dict_add(meta,
+                                 pmt::mp("sample_offset"),
+                                 pmt::from_uint64(nitems_read(0) + i));
             message_port_pub(
                 pmt::mp("out"),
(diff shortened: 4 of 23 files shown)


daniestevez:
Can you describe the approach you have used here? The challenge in implementing this correctly is that the Symbol Sync block (which is present in most receiver chains) does variable resampling to try to synchronize to the transmitter symbol clock. Therefore, counting samples at the output of Symbol Sync is not an appropriate way to count time, because the sample rate at that point is variable. A mechanism to trace back what input samples correspond to Symbol Sync output samples is needed.

parker-research:
Good question! I think I'm still confident in the approach being a solid improvement, but I agree that it's not perfect.

The tl;dr is that you're totally right that it tracks symbols instead of samples, but in practice, they're very strongly aligned in timing, and thus are likely good enough, at least for the first revision here. 

### Implementation Details

The `sample_offset` metadata that some deframers attach (fixedlen_to_pdu, lilacsat1_demux, distributed_syncframe_soft, and the HDLC/AX.25 deframer) is a symbol count, taken at the deframer's own input port, which is downstream of Symbol Sync. `pdu_add_timestamp` (totally new in this PR) converts that count to an absolute timestamp via `start_time + sample_offset / baudrate`, using the nominal baudrate from the SatYAML file.

You're right that this isn't a sample-exact backtrace (and is instead "symbol-exact"). As I'm sure you obviously know, Symbol Sync's output is a variable-rate resampler that continuously adjusts its interpolation ratio to track the recovered symbol clock, so "symbols counted" and "elapsed input-sample time" aren't rigidly 1:1. I don't have a mechanism that traces output symbols back to the corresponding input samples (at least not yet).

That said, I think the resulting error is small enough to be a good tradeoff in practice, and it's bounded by the implementation from before this PR: every demodulator constructs `symbol_sync_cc` with `clk_limit = self.options.clk_limit * sps`, where `clk_limit` defaults to 0.004 (0.4%). That's the hard ceiling on how far the recovered symbol rate is corrected. Obviously, it's the maximum the loop can lock onto, and not a typical value. So:

- Worst-case (pathological) drift is bounded at 0.4% of elapsed time (e.g. <=2.4s over 10-minutes, and normally way less).
- Real transmitter/receiver oscillator errors are typically tens of ppm, i.e. two to three orders of magnitude tighter than that ceiling, so in practice the drift is millisecond-scale over a full pass.
- The loop's frame-to-frame timing jitter doesn't accumulate. Instead, it only oscillates around the true phase within a fraction of a symbol period.

So this isn't as rigorous as tracing back through the resampler, but it should be accurate to well under a symbol period in the vast majority of real recordings, and it's a large improvement over the previous behavior (which required exact 1x `--throttle`ed playback to avoid errors of the full recording duration).


### Next Steps

I'm open to a tag-propagation-based approach that traces back to input samples if you'd prefer that level of rigor, but wanted to check whether this approximation is acceptable first before investing in that.

I'd suggest letting this get merged and released, and then we'll improve on it further in a separate PR. Things don't have to be perfect to be better.

daniestevez:
Thanks for the reply. Since I'm in the process of thinking what to write in a future AI policy for gr-satellites (because we will need to have one at some point), let me ask you the following questions:

- Is the code in this PR generated with an LLM?
- Is the bulk of your previous comment generated with an LLM?
- Would that hypothetical PR that traces back to input samples be implemented by an LLM?
- If the answer to any of these questions is 'yes', then what is the reason for using an LLM instead of doing the work manually? Is it to save time? Is it to obtain technical help from the LLM (that is, to make the LLM handle technical details about gr-satellites and GNU Radio that the programmer might not know)? Is it for another reason (if so specify)?

parker-research:
Hey! Happy to help answer and cooperate.

#### Your Questions

> Is the code in this PR generated with an LLM?

Yes, some of it is partially generated with an LLM. As you may know, LLMs kinda suck often, and so, parts are hand-written, hand-edited, etc.

> Is the bulk of your previous comment generated with an LLM?

The Implementation Details part is perhaps 80% LLM-generated (and then reviewed, editing, and believe-it-or-not stripped down). Iirc, the rest is hand-written.

> Would that hypothetical PR that traces back to input samples be implemented by an LLM?

Yes, I'll likely use an LLM to help map out the data flow path from samples to symbols, and possibly to help to instrument a v1 of the code.

As was the case with this PR, it'll be reviewed and tested manually, and any tradeoffs will be evaluated with my judgement.

> If the answer to any of these questions is 'yes', then what is the reason for using an LLM instead of doing the work manually? Is it to save time? Is it to obtain technical help from the LLM (that is, to make the LLM handle technical details about gr-satellites and GNU Radio that the programmer might not know)? Is it for another reason (if so specify)?

My research focus and goals in life aren't really in modifying hybrid C++/Python projects for decoding satellite packets; instead, this is just a means to an end of getting the satellite packets. In that regard, saving time and reaching the testable end result was the main goal.

As an experienced software developer, I have no doubt I'd have reached a very similar solution/implementation, given 10-40x more time. 

The LLM was certainly helpful for figuring out the data flow path (at least 80% of it, hence the discussion), and for figuring out the pybind11/C++-Python interop parts. I've written a good amount of that interop code by hand, and it's definitely my least-favourite type of thing to work on. 

I could've just made the change and used it in private. Or, in a horribly-increasing trend in open source software development, I could've told an LLM to rebuild this whole project from scratch in a language I prefer, based on this project, for the demoduation types I desire. I believe what I've done here is the far-more-responsible approach in the LLM-era, in that it doesn't shard the community.

#### Next Steps

Please advise where you'd like to go from here! Happy to close, happy to edit, or anything else!

daniestevez:
Thanks for your detailed reply and your transparency. Some comments to things you mentioned:

> My research focus and goals in life aren't really in modifying hybrid C++/Python projects for decoding satellite packets; instead, this is just a means to an end of getting the satellite packets. In that regard, saving time and reaching the testable end result was the main goal.

This makes a lot of sense. However, gr-satellites, as an open source software project about decoding satellite packets using C++/Python is more about the code itself than about the end result only. It is understandable that many users only care about the end result (that is, what the software can do for them), but as a maintainer I need to balance this with the longer term sustainability of the project and its technical quality.

> As an experienced software developer, I have no doubt I'd have reached a very similar solution/implementation, given 10-40x more time.

This also makes sense to me. It makes sense that a developer/user only interested in results would prefer to spend less time solving this problem. However, the [Rust LLM policy](https://forge.rust-lang.org/policies/llm-usage.html) says: "LLMs work best when used as a tool to write better, not faster." I am aligned with that thought, and something like this will probably go into gr-satellite's policy eventually.

> and for figuring out the pybind11/C++-Python interop parts.

I don't follow. Your PR doesn't touch any pybind11 code, and generally `gr_modtool` can regenerate the pybind11 bindings just fine (the main exception is when templates are involved). Your C++ changes only modify some blocks to pass along the `car` of a PDU (which is something that these blocks should be doing anyway, and I would be interested in getting those changes in in any case), and to create a `car` containing the `sample_offset` in some other blocks. This does not interact with Python in any direct way. Understanding this part of the changes only requires understanding how PDUs are used in GNU Radio.

> I could've just made the change and used it in private. Or, in a horribly-increasing trend in open source software development, I could've told an LLM to rebuild this whole project from scratch in a language I prefer, based on this project, for the demoduation types I desire. I believe what I've done here is the far-more-responsible approach in the LLM-era, in that it doesn't shard the community.

This also makes sense, and is appreciated. However the situation here is that by trying to upstream this code, you are requesting a maintainer to review it and to maintain the code long term, owning both solving any future problems that stream from these changes and having to make changes to this code when required by refactors elsewhere in the code base. The effort required to review the PR properly (which is something that a human needs to do in any case) can be comparable to that of getting an LLM to write the PR in the first place.

After I wrote #92, I came up in my mind with an idea about how the right solution to this problem would look like, but as I pointed out in the issue, the right solution is not trivial. As this issue has never been a priority, I have never allocated time to solving it. If I had someone interested on implementing this feature for the sake of contributing to gr-satellites with high quality code that solves a difficult problem in a good way, I would be more than willing to invest time of my own in supporting this by providing advice and reviewing and testing the PR. Such kind of contributor would also be more likely to stick around and collaborate with project further. However, in your case you say that your goal is to get the satellite decoder features you need and to not invest much time into this. It seems that the effort commitment you want to do is not even enough to manually type from scratch a description about how your approach works when I asked for it. So here is where the disconnect is.

I think it's great that you've found a way to make gr-satellites do what you need without investing too much effort (that is basically in line with the first part of freedom 3 in the definition of free software). I also think it's great that you've made available your changes to others and thought of upstreaming them (that is in line with the second part of freedom 3 in the definition of free software). However, that doesn't mean that I'm willing to get merged either these changes or the potential future PR that follows the approach I outlined.

My proposal here is just to close this PR. Your changes can live in your fork of gr-satellites (as they already live), and perhaps someone else might also find them useful. I hope that you understand my stance. It is nothing personal and nothing against LLMs (I also use them frequently at work).

parker-research:
I promise we don't have dissimilar views on LLMs. Thanks for engaging.

Regarding review effort, I totally agree, but I do mean to highlight that I've kept this PR extremely small. This is an isolated, 120 SLOC (+ comments) change.

I'm interested in getting this change (or a change that aligns what you want to see in your project) merged. Happy to even write it by hand, if that's a sticking point. Can you please advise in the #92 issue what your envisioned solution looks like - you mentioned you have a concept that this doesn't align with.

parker-research:
Closing as requested - arguing with strangers on the internet isn't _that_ much fun.

daniestevez:
@parker-research I have explained in #92 the vision I have for the solution, since in any case I think that will be good context for anyone running across this in the future.

The solution you have in this PR can be acceptable in some cases, but in other cases such as leaving the receiver running on AWGN for tens of minutes with a Symbol Sync that has a relatively large Max Deviation will produce timestamps which are pretty far off.

PR #848, opened 2026-08-18
Add FrontierSat satyaml file

parker-research:
To validate, you can download the audio from https://network.satnogs.org/observations/14813295/ (or any observation). For example:

```bash
wget 'https://network-satnogs.freetls.fastly.net/media/data_obs/2026/8/18/17/14813295/satnogs_14813295_2026-08-18T17-05-35.ogg'
sox satnogs_14813295_2026-08-18T17-05-35.ogg satnogs_14813295_2026-08-18T17-05-35.wav
gr_satellites python/satyaml/FrontierSat.yml --hexdump --wavfile satnogs_14813295_2026-08-18T17-05-35.wav
```

### DIFF
--- a/python/satyaml/FrontierSat.yml
+++ b/python/satyaml/FrontierSat.yml
@@ -2,7 +2,7 @@ name: FRONTIERSAT
 norad: 69015
 data:
   &tlm Telemetry:
-    telemetry: hexdump
+    unknown
 transmitters:
   9k6 FSK downlink:
     frequency: 436.150e+6
--- a/python/satyaml/FrontierSat.yml
+++ b/python/satyaml/FrontierSat.yml
@@ -1,0 +1,17 @@
+name: FRONTIERSAT
+norad: 69015
+data:
+  &tlm Telemetry:
+    telemetry: hexdump
+transmitters:
+  9k6 FSK downlink:
+    frequency: 436.150e+6
+    modulation: FSK
+    baudrate: 9600
+    deviation: 3200
+    framing: AX100 ASM+Golay
+    scrambler: CCSDS
+    syncword: 4
+    data:
+    - *tlm
+
--- a/python/satyaml/FrontierSat.yml
+++ b/python/satyaml/FrontierSat.yml
@@ -10,8 +10,8 @@ transmitters:
     baudrate: 9600
     deviation: 3200
     framing: AX100 ASM+Golay
-    scrambler: CCSDS
-    syncword: 4
     data:
     - *tlm
+
+# AX100 Configuration: https://github.com/CalgaryToSpace/CTS-SAT-1-Ground-Station/blob/main/docs/AX100_Config_Dump.txt
 


python/satyaml/FrontierSat.yml
```diff
@@ -0,0 +1,17 @@
...
+  &tlm Telemetry:
+    telemetry: hexdump
+transmitters:
+  9k6 FSK downlink:
+    frequency: 436.150e+6
+    modulation: FSK
+    baudrate: 9600
+    deviation: 3200
+    framing: AX100 ASM+Golay
+    scrambler: CCSDS
```
daniestevez: This is the default for the AX100 ASM+Golay deframer, so I suggest to leave it out, as is done in similar SatYAML files.

python/satyaml/FrontierSat.yml
```diff
@@ -0,0 +1,17 @@
...
+    syncword: 4
```
daniestevez: Why is this field included? I don't think anything actually reads it.

python/satyaml/FrontierSat.yml
```diff
@@ -0,0 +1,17 @@
+name: FRONTIERSAT
+norad: 69015
+data:
+  &tlm Telemetry:
+    telemetry: hexdump
```
daniestevez:
```suggestion
    unknown
```

When there is not an appropriate telemetry parser, we simply use the `unknown` keyword, like this: https://github.com/daniestevez/gr-satellites/blob/main/python/satyaml/239Alferov.yml#L5

python/satyaml/FrontierSat.yml
```diff
@@ -0,0 +1,17 @@
...
+transmitters:
+  9k6 FSK downlink:
+    frequency: 436.150e+6
+    modulation: FSK
+    baudrate: 9600
+    deviation: 3200
```
daniestevez, line 11: This deviation is 1/3 of the symbol rate, which is an uncommon choice (MSK would be 1/4 of the symbol rate). Is this deviation stated in public documentation or has it been measured on IQ recordings from the satellite?

daniestevez (changes requested):
Thanks for adding this! A few comments.

python/satyaml/FrontierSat.yml
parker-research: Mostly for safety and to be explicit! Defaults can change over time (though this one likely won't, of course).

python/satyaml/FrontierSat.yml
parker-research: Just a mistake - SatYAML specification is a little lacking.

python/satyaml/FrontierSat.yml
parker-research: The satellite's AX100 config dump is here: https://github.com/CalgaryToSpace/CTS-SAT-1-Ground-Station/blob/main/docs/AX100_Config_Dump.txt

Correct me if I'm wrong in any of this understanding, but my understanding is that the relevant parameter is `modindex`, which is set to 0, which means "use the default". Per the AX100 datasheet, it gives this logic definition (assuming `modindex` and `modfactor` refer to the same thing):

```c
if (modfactor == 0) {
   if (baudrate < 1300) {
      modfactor = 1 / 1.2;
   } else if (baudrate < 60000) {
      modfactor = 1 / 1.5;
   } else {
      modfactor = 1 / 2;
   }
}
```

Thus, `modfactor = 1 / 1.5` = 0.666667.

Then, we use the equation: `F_dev = (baudrate * modfactor) / 2`

Thus, `F_dev = (9600 * 0.666667) / 2 = 3200 Hz`

Happy to share an IQ recording if you're interested in (or perhaps willing to) dig into it further.

parker-research (commented):
Updated review comments. Looking forward to hearing your thoughts on the frequency deviation.

python/satyaml/FrontierSat.yml
daniestevez: This sounds reasonable, but I'd like to verify this in the IQ recording. Let's get this PR merged, I'll look at the IQ recording when you share it, and if changes are needed, I can get those into `main` later.

daniestevez (approved):

parker-research:
Here's two IQ files shared from another SatNOGS station owner (from debugging another SatNOGS issue). Will share these as that's easiest at the moment. https://pe0sat.vgnet.nl/download/FRONTIERSAT/

daniestevez:
Thanks! This is a plot of the waveform from a packet in the first recording, which has been demodulated with the expected 3200 Hz deviation. This confirms that the deviation is indeed 3200 Hz.

waveform_packet

parker-research:
Neat! Thanks for sharing that! First time I've seen such a plot

daniestevez:
There is a [small tool](https://github.com/daniestevez/gr-satellites/blob/main/tools/waveform_plot.py) to make these plots (but the user needs to select the data corresponding to a packet manually).

PR #852, opened 2026-09-09
Support GR 3.11 CMake breaking changes

bgottula:
In commit [24e9f2b](https://github.com/gnuradio/gnuradio/commit/24e9f2b7ef19806ce7754bdfc6622ed17355454) on GNU Radio `main`, `GR_LIBRARY_FOO` became a [deprecated alias](https://github.com/gnuradio/gnuradio/blob/24e9f2b7ef19806ce7754bdfc6622ed17355454c/cmake/Modules/GrMiscUtils.cmake#L125-L126) for `GR_REGISTER_LIBRARY` which only registers the target but does not install it. Install rules have to be requested explicitly via the new `GR_INSTALL_LIBRARY`. This is an intentional compatibility-breaking change anticipated for the 3.11 release, as mentioned at GRCon 2025 (slide 15 of the [Maintainer's Update](https://events.gnuradio.org/event/26/contributions/796/attachments/251/653/presentation.pdf)).

This PR adds support for the new commands when they exist and falls back to `GR_LIBRARY_FOO` for backwards compatibility with GNU Radio <3.11.

This has been tested by building gr-satellites against both the released GNU Radio 3.10.12.0 conda-forge package and against commit aee9fd3 (tip of main as of 9 Sept 2026) on Ubuntu 24.04 and confirming that all QA tests pass.


daniestevez (approved):
Excellent! This looks good. Merging. Let me know if you run into any other issues with gr-satellites on GR 3.11, since we definitely want to support a smooth update when 3.11 comes out.

Issue #849, opened 2026-08-19
Request: Add an AI policy, maybe


parker-research:
Per the discussion in #846, it'd be helpful to have an AI usage policy for contributions in this repository.

daniestevez:
I agree this is greatly needed. I have been thinking of writing a specific policy for gr-satellites (rather than adopting a policy used by a larger project), however I'm starting to come to the realization that this would be a lot of work and very difficult to get right. So far I'm tempted to adopt the policy recently created by the Rust project ([policy](https://forge.rust-lang.org/policies/llm-usage.html) and [announcement](https://blog.rust-lang.org/inside-rust/2026/08/05/rust-langrust-is-adopting-an-llm-policy/)), striking out the parts which are too specific to how the Rust project is run (e.g., Zulip channels), but I might do a wider survey to see if there is anything out there that might be a better fit. I wanted to take the opportunity of using this issue to float this idea around to see if there are people who have strong thoughts.

xdsopl:
Some projects have a no-AI policy now but that feels like denying yourself some great contributions. I only accept contributions when the author can explain every single change. That means every commit needs to be small and has a meaningful message. Then it doesn't mater if it was AI generated.

daniestevez:
I have merged #855, where an LLM policy that is essentially the same as the Rust project's policy (with minor modifications where sentences refer to technical elements that are exclusive to the Rust project). I have been delaying this too much, partly to try to find a better alternative, but we now need an LLM policy in place, partly to know what to do with PRs like #853.

I will close this issue, because the LLM policy is now added, but anyone should feel free to reopen it or to open a Github discussion to give feedback about the policy.

PR #853, opened 2026-09-10
Add plaintext telemetry decoding for Geoscan 1-6, 239Alferov and GRBBeta

F1EBK:
Display telemetry of Geoscan-1 to Geoscan-6, 239Alferov and GRBBeta in a human readable form  :

 -> Packet from 9k6 FSK downlink
address = RS92S2 -> BEACON
time = 2026-09-09T03:28:42+00:00
eps_mode = 3
current_total = 200 mA
current_solar_panels = 210 mA
... 

 New telemetry decoder modules: geoscan_beacon.py and grbbeta.py
   Updated the corresponding SatYAML files and telemetry/__init__.py accordingly.

telemetry structure are from : 
-  a public definition for GRBBeta
-  the geoscan telemetry format definition in the SkyRoof software (open source software) with the acknowledge of VE3NEA.


F1EBK:
fixing picodestyle issue

F1EBK:
The build-ubuntu-gr310 check fails on this PR with:
  fatal: detected dubious ownership in repository at '/github/workspace'
which then seems to cascade into ModuleNotFoundError: No module named
'satellites' on all 27 remaining tests.

This looks specific to the fork-based pull_request context: commit
cfa6aac ("Support GR 3.11 library installation CMake"), which is already
included in this branch, passes build-ubuntu-gr310 fine when run as a
direct push on main (run #1971), but fails here. This suggests the CI
workflow may need `git config --global --add safe.directory` added for
the fork/pull_request checkout context.

daniestevez:
I want to be fully transparent with this PR, especially for anyone seeing this without context.

On 2026-09-06, @F1EBK emailed me saying that he had used an AI to build a Python program to parse GRBBeta telemetry. He asked whether this could be integrated in gr-satellites. My full reply was as follows:

> Thanks for asking.
>
> If the telemetry format is publicly documented, you are willing to adapt this code to make it a gr-satellites telemetry definition like the ones already in tree (using construct, etc.), and you can be accountable during the PR review process (even if an AI wrote it, it is your code, you have read it, you understand it, etc.), then I'm interested in getting this merged in gr-satellites.
>
> If on the other hand you simply want to put this code out there with little additional effort, then gr-satellites is not the right place, but there are many places where you can put this code to allow others to use it.

Then he mentioned that he was not familiar with Github,  and that he did not know what I was talking about regarding construct. He asked whether there are howtos or videos that would help him understand how to contribute. Here is my full reply to that:

> I don't know of a howto or video, but I'm pretty sure that there is lots of information online about how to use git and Github, and how to collaborate in open source development.
> 
> Regarding construct, etc., read the gr-satellites code and construct's documentation! Contributing to a code base that you don't understand can lead to a frustrating experience.

We haven't exchanged any further emails.

For a while now I have thought that gr-satellites needs an LLM usage policy (see #849 and discussions that lead to that), and this PR is what has pushed me to adapt Rust's policy and adopt it in gr-satellites (#855). I believe that without an LLM usage policy, this PR and similar PRs would likely lead to a frustrating experience for everyone.

@F1EBK, on the technical aspects of this PR:

- CI fails for this PR, as you have noticed. You have posted a comment (I don't know if this is you or just a copy & paste of an LLM response) saying that the problem is #852, a PR that was merged recently. I don't think that's the case. CI passes in `main` and also in other PRs that have been opened since that PR was merged. It is very likely that it is the changes in your PR the ones which are breaking CI. At a first glance it's not obvious to me why they break CI, though. Your comment says that the CI workflow might need `git config --global --add safe.directory`, which makes no sense without further justification. That is generally only needed when there are file ownership problems (files owned by a user different than the one that is running the build), and the errors that these problems cause are usually different from the error we are seeing here.

- One of your files contains a huge chunk of comments in French. Have you read all the code you have submitted in the PR? Did you really intend to submit comments in French? I should clarify that I don't have anything against non-English languages (English is my second language, in fact), but everything in gr-satellites is in English, for the obvious reason that it is the de facto standard for international technical work.

- In my email I asked about public documentation for the telemetry formats. That is the main way of checking if the code is correct. You mention something about public documentation in the PR description, but provide no links. Surely we can Google, but it would be better to reference to the same documentation you have used to build this. Additionally, the French comment references a local file as the source of truth (`%APPDATA%\\Afreet\\Products\\SkyRoof\\TelemetryRegistry\\geoscan2.json`). The PR description says "the geoscan telemetry format definition in the SkyRoof software (open source software) with the acknowledge of VE3NEA". [SkyRoof](https://github.com/VE3NEA/SkyRoof) is licensed as AGPL-3.0, which is incompatible with the GPL-3.0-or-later licensing used by gr-satellites. We cannot merge code from SkyRoof into gr-satellites, regardless of whether that code has been massaged by an LLM. Admitting that the code comes from SkyRoof is very likely to make it a derivative work.

- Your Geoscan code doesn't follow the typical architecture of gr-satellites telemetry parsers, which is what I asked for in my email. As you can see in the code, almost all the gr-satellites telemetry parsers are just a series of construct definitions (which get hierarchically embedded into each other). You have implemented a class that contains a couple construct definitions but implements its own `parse()` method that returns a formatted string rather than an object.

- I could say the same for the GRBBeta implementation, but since referring you to construct, I did my homework and looked at [their webpage](https://grbbeta.tuke.sk/index.php/en/home/) and saw that their beacons are basically a CSV ASCII string. construct is not the right tool to parse that, so we would need custom code for the parsing, but probably we still want to return an object rather than a formatted string.

I know that you have opened this PR before I added the LLM policy (in fact this PR is what finally pushed me to add the LLM policy). However the fact that an LLM policy was incoming was not a secret. There was an issue open about it and I also asked for feedback on all my social networks. So this should not come as a surprise. I hope that the bullet points I've listed have convinced you that this PR is currently not in the right shape that would convince a reviewer that the review process is going to be fruitful. My suggestion is that you close this PR, and if you still want to upstream this, make a new PR following the LLM policy and also being mindful about taking code from projects whose license is not compatible with gr-satellites.

F1EBK:
I made this PR as you told me "you were interested in getting this merged in gr-satellites". I told you that I was a real newbie in GitHub. So keep this in mind. No problem about the comments in french, I can translate it in english as it is also my second language (it won't be AI generated). Now If this doesn't fit the typical architechture of gr_satellites, I can't help it. I made this piece of software to help  others getting humain readable informations out of a set of bytes transmited by some satellites. Keep in mind that I am also a newbie in gr_satellites and never wanted to bug your software nor others.  I would have hope-for a little help from you, and not a simple reject. So I give up in this process, and won't do any more PR. 73.

PR #854, opened 2026-09-10
satyaml: add GALAPAGOS-UTE-SWSU

reynico:
Hi Dani, I confirm that this configuration decodes GALAPAGOS's telemetry

additional context: https://community.libre.space/t/galapagos-ute-swsu-transmissions/14209/20


python/satyaml/GALAPAGOS-UTE-SWSU.yml
```diff
@@ -0,0 +1,18 @@
...
+data:
+  &tlm AX25 telemetry:
+    telemetry: ax25
+transmitters:
+  1k143 AFSK downlink:
+    frequency: 437.050e+6
+    modulation: AFSK
+    baudrate: 1143
+    af_carrier: 1714
+    deviation: 571
```
daniestevez, line 14: Why these values? They are close to standard 1k2 AFSK that uses Bell 202 tones at 1200 and 2200 Hz (that would correspond to baudrate 1200, af_carrier 1700, deviation 500). Where do these values come from? Have they been measured accurately?

reynico: hi Dani, I measured them from a Satnogs recording: https://network.satnogs.org/observations/14959822/

- FFT shows two tones 1142.5-1143.3 Hz and 2285.0-2286.7 Hz 
- af_carrier/deviation in the yml are just the mean of those measured tones: (1142.9+2285.8)/2 = 1714, (2285.8-1142.9)/2 = 571.

I validated the values by de-framing the recording and got AX.25 with dest ALL, src HC1PX, PID 0xF0, etc.

daniestevez: Where does the baudrate measurement come from? The frequency of the AFSK tones is independent from this.

reynico: the baudrate was already configured in the SatNOGS database for that transmitter: https://db.satnogs.org/satellite/XHOE-9718-5112-5196-7058#transmitters

daniestevez: Okay. It's interesting, because these are weird numbers. It almost looks like they attempted to do standard 1k2 AFSK but had some frequency errors in the implementation. Let's get this merged as is, then.

daniestevez (approved):

